PR #8, opened 2024-07-05
Set up CI including packaging and publishing to NuGet

dslmeinte:
(I'd prefer to squash-merge in the end)


src/LionWeb-CSharp-Generator/LionWeb-CSharp-Generator.csproj
```diff
@@ -15,4 +21,15 @@
         <PackageReference Include="Microsoft.CodeAnalysis" />
     </ItemGroup>
 
+    <ItemGroup>
+      <None Update="LICENSE">
```
enikao, line 25: This is a rather interesting tag name ... (nothing to be done here, I'm just baffled by the name).

enikao (approved):

PR #9, opened 2024-07-08
factory interface generator

enikao:


dslmeinte (approved):

JojoEffect:
Suggestion: If the generated XYZFactories would now additionally provide "virtual" implementations - one could partially override the factories and reuse the rest of them as implementation for the IXYZFactory. What do you think @enikao?

enikao:
> Suggestion: If the generated XYZFactories would now additionally provide "virtual" implementations - one could partially override the factories and reuse the rest of them as implementation for the IXYZFactory. What do you think @enikao?

I thought delegation is so much better than specialization (-:
Sure, can be done.
I don't really get the use case, though -- what would such a specialization do different? We can already override `GetNewId()`.
Changing e.g. `CreateCircle()` to not only create the circle, but also add some default center coordinate, would break the semantics of the method -- I'm not sure how the rest of the code would react to that.

JojoEffect:
> > Suggestion: If the generated XYZFactories would now additionally provide "virtual" implementations - one could partially override the factories and reuse the rest of them as implementation for the IXYZFactory. What do you think @enikao?
> 
> I thought delegation is so much better than specialization (-: Sure, can be done. I don't really get the use case, though -- what would such a specialization do different? We can already override `GetNewId()`. Changing e.g. `CreateCircle()` to not only create the circle, but also add some default center coordinate, would break the semantics of the method -- I'm not sure how the rest of the code would react to that.

In this case we will get both with the virtual keyword. We can delegate a subset of the interface implementation to a new implementation and keep the rest of the implementation. Of course one could also do this by composition and delegating the unchanged calls to a instence of the generated factory.

I'm not convinced that a Circle with a default center Coordinate breaks the semantics in this sense. At least not in the case when the center Coordinate is required and not optional. Therefore I would state it the other way araound and would say that the generated default implementation of the `CreateCircle()` method is incomplete and does not fulfill its suggested behaviour. (I still know and agree with the reasons for this API design)

enikao:
Ok, makes sense. Curious to see the effects.

JojoEffect:
> Ok, makes sense. Curious to see the effects.

Do you expect any missuse of that API?

enikao:
I'm not sure, I can't oversee it. That's why I'm curious (-:

JojoEffect:
Well, regarding this I'm almost all the time curious 😅

enikao:
@JojoEffect implemented, please have a look

JojoEffect (approved):
Thanks :-)

Issue #7, opened 2024-07-05
Textualizer: LionWeb.Core.UnsetFeatureException


erk-dik:
`Textualizer.AsString(node); ` throws the following exception:

```
  LionWeb.Core.UnsetFeatureException
  Required Property name is not set
...
   at LionWeb.Core.Utilities.Textualizer.<AsStrings>g__ReferenceTargetAsString|4_2(INode target)
   at System.Linq.Enumerable.SelectIListIterator`2.Fill(IList`1 source, Span`1 results, Func`2 func)
   at System.Linq.Enumerable.SelectIListIterator`2.ToArray()
   at LionWeb.Core.Utilities.Textualizer.Join(IEnumerable`1 strings, String separator)
   at LionWeb.Core.Utilities.Textualizer.<>c__DisplayClass4_0.<AsStrings>g__ReferenceAsString|3(Reference reference)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.List`1.AddRange(IEnumerable`1 collection)
   at LionWeb.Core.Utilities.Textualizer.AsStrings(INode node)
   at LionWeb.Core.Utilities.Textualizer.AsString(INode node)
...
```

dslmeinte:
@erk-dik Can you implement a failing unit test for this somewhere?

dslmeinte:
@erk-dik @JojoEffect Should be fixed by PR #10 .

dslmeinte:
@erk-dik Can I make you collaborator of this repo? Then I can request reviews of PRs from you.

erk-dik:
Sure  @dslmeinte

dslmeinte:
> Sure @dslmeinte

Invite sent :)

PR #10, opened 2024-07-08
Fix textualizer

dslmeinte:
It throws on references to named nodes having an unset name — this fixes issue #7.

### DIFF
--- a/src/LionWeb-CSharp/core/Utilities/Textualizer.cs
+++ b/src/LionWeb-CSharp/core/Utilities/Textualizer.cs
@@ -41,9 +41,11 @@ public static class Textualizer
 
     private static IEnumerable<string> AsStrings(this INode node)
     {
+        var noNameSetOutput = "<no name set!>";
+
         return
         [
-            $"{node.GetClassifier().Name} (id: {node.GetId()}) {{",
+            $"{Name(node.GetClassifier())} (id: {node.GetId()}) {{",
             ..Indent([
                 ..node.CollectAllSetFeatures().OfType<Property>().Select(PropertyAsString),
                 ..node.CollectAllSetFeatures().OfType<Reference>().Select(ReferenceAsString),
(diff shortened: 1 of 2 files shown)


src/LionWeb-CSharp/core/Utilities/Textualizer.cs
```diff
@@ -63,8 +63,20 @@ string ValueAsString(object? value)
         string PropertyAsString(Property property)
             => $"{property.Name} = {ValueAsString(node.Get(property))}";
```
enikao, line 64: This would also crash if `property` is unset and required.

src/LionWeb-CSharp/core/Utilities/Textualizer.cs
```diff
@@ -63,8 +63,20 @@ string ValueAsString(object? value)
...
+        string NameInfo(INode node)
+        {
+            if (node is INamed named)
+            {
+                return named.CollectAllSetFeatures().Contains(BuiltInsLanguage.Instance.INamed_name)
+                    ? $"({named.Name})"
+                    : "<no name set!>";
+            }
+
+            return "<not named>";
```
enikao, line 75: Is this useful output or clutter?

src/LionWeb-CSharp/core/Utilities/Textualizer.cs
enikao: same for `ReferenceAsString()` and `ContainmentAsString()`

dslmeinte: That won't happen (in the same way) as those `Name` properties are `required`.

src/LionWeb-CSharp/core/Utilities/Textualizer.cs
dslmeinte: Maybe the `<not named>` is clutter, but if you'd expect a reference target to have a name, then being informed that the textualizer also thinks it should have one, would be useful to me.

src/LionWeb-CSharp/core/Utilities/Textualizer.cs
enikao: Do you have a test for that?

src/LionWeb-CSharp/core/Utilities/Textualizer.cs
enikao: Does the textualizer expect _all_ reference targets to have a name?

dslmeinte: Uh no, which is why I'd be willing to agree that the `<not named>` is clutter.

src/LionWeb-CSharp/core/Utilities/Textualizer.cs
dslmeinte: I'm guessing you're asking because you know that if I had one, it would fail :D

Issue #14, opened 2024-07-09
CI should trigger on PR


dslmeinte:

enikao:
... and prevent the PR from being merged if the pipeline fails.

dslmeinte:
Fixed by PR #16

PR #15, opened 2024-07-09
Trigger CI on PR

dslmeinte:
Issue #14

### DIFF
--- a/test/LionWeb-CSharp-Test/StopCI.cs
+++ b/test/LionWeb-CSharp-Test/StopCI.cs
@@ -24,7 +24,7 @@ public class StopCI
     [TestMethod]
     public void stop_CI_from_publishing_to_NuGet_org()
     {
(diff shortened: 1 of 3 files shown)


test/LionWeb-CSharp-Test/StopCI.cs
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// 
+// SPDX-FileCopyrightText: 2024 TRUMPF Laser SE and other contributors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace LionWeb_CSharp_Test;
+
+[TestClass]
+public class StopCI
```
enikao, line 21: This feels like a weird way to stop publishing. Can't we have two different actions, or different triggers?

dslmeinte: Just for testing...please ignore for now

PR #17, opened 2024-07-09
Implement naive reference finder

dslmeinte:
+ update changelog for issue #7


src/LionWeb-CSharp/core/Utilities/ReferenceExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+// ReSharper disable NotAccessedPositionalProperty.Global
+public record ReferenceValue(IReadableNode SourceNode, Reference Reference, int? Index, IReadableNode TargetNode);
+
+/// <summary>
+/// Extension methods to deal with references, i.e. values of <see cref="Reference"/> features.
+/// </summary>
+public static class ReferenceExtensions
+{
+    private static IEnumerable<ReferenceValue> GatherReferenceValues(IReadableNode sourceNode, Reference reference)
+        => reference.Multiple
```
enikao, line 41: I would write this as statements, not one huge expression. We have more than 3 branches in here.

src/LionWeb-CSharp/core/Utilities/ReferenceExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+            ? (sourceNode.Get(reference) as IEnumerable<IReadableNode> ?? []).Select((targetNode, index) =>
+                new ReferenceValue(sourceNode, reference, index, targetNode))
+            : [new ReferenceValue(sourceNode, reference, null, (sourceNode.Get(reference) as IReadableNode)!)];
+
+    /// <summary>
+    /// Finds all references within the given <paramref name="scope"/>, as <see cref="ReferenceValue"/>s.
+    /// To search within all nodes under a collection of root nodes,
+    /// pass <code>rootNodes.SelectMany(rootNode => rootNode.Descendants(true, true)</code> as scope.
```
enikao, line 49: use `<c>` tags for inline code examples. `code` creates a block.

src/LionWeb-CSharp/core/Utilities/ReferenceExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// </summary>
+    /// <param name="scope">The <see cref="IReadableNode"/>s that are searched for references</param>
+    /// <returns>An enumeration of references, as <see cref="ReferenceValue"/>s.</returns>
+    public static IEnumerable<ReferenceValue> ReferenceValues(IEnumerable<IReadableNode> scope)
+        => scope
+            .SelectMany(sourceNode =>                                                       // for all nodes in the scope:
+                sourceNode
+                    .CollectAllSetFeatures().OfType<Reference>()                            // for all set references:
```
enikao, line 57: why `OfType()` on the same line?

src/LionWeb-CSharp/core/Utilities/ReferenceExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+                    .SelectMany(reference => GatherReferenceValues(sourceNode, reference))  // gather all reference values
+            );
+
+    /// <summary>
+    /// Finds all references coming into any of the given <paramref name="targetNodes"/>
+    /// within the given <paramref name="scope"/>, as <see cref="ReferenceValue"/>s.
+    /// To search within all nodes under a collection of root nodes,
+    /// pass <code>rootNodes.SelectMany(rootNode => rootNode.Descendants(true, true)</code> as scope.
```
enikao, line 65: `<c>` tag

src/LionWeb-CSharp/core/Utilities/ReferenceExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+        var targetNodesAsSet = new HashSet<IReadableNode>(targetNodes);
+        return ReferenceValues(scope)
+            .Where(referenceValue => targetNodesAsSet.Contains(referenceValue.TargetNode));
+    }
+
+    /// <summary>
+    /// Finds all references coming into the given <paramref name="targetNode"/>
+    /// within the given <paramref name="scope"/>, as <see cref="ReferenceValue"/>s.
+    /// To search within all nodes under a collection of root nodes,
+    /// pass <code>rootNodes.SelectMany(rootNode => rootNode.Descendants(true, true)</code> as scope.
```
enikao, line 82: `<c>` tag

src/LionWeb-CSharp/core/Utilities/ReferenceExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+/// or <code>null</code> if the reference feature is not multivalued
+/// <param name="TargetNode">The target <see cref="IReadableNode"/> of the reference (relation)</param>
+/// </param>
+// ReSharper disable NotAccessedPositionalProperty.Global
+public record ReferenceValue(IReadableNode SourceNode, Reference Reference, int? Index, IReadableNode TargetNode);
+
+/// <summary>
+/// Extension methods to deal with references, i.e. values of <see cref="Reference"/> features.
+/// </summary>
+public static class ReferenceExtensions
```
enikao, line 38: Is this a good name, given there is no extension method in here?

test/LionWeb-CSharp-Test/tests/ReferenceExtensionsTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        var refs = ReferenceExtensions.FindIncomingReferences(geometry.Shapes[0], rootNodes).ToList();
+        Assert.AreEqual(1, refs.Count());
+        var ref0 = refs.First();
+
+        var expectedReferenceValue =
+            new ReferenceValue(referenceGeometry, language.ReferenceGeometry_shapes, 0, geometry.Shapes[0]);
+        Assert.AreEqual(expectedReferenceValue, ref0);  // (relies on value equality of C# record types)
+
+        var allRefs = ReferenceExtensions.ReferenceValues(rootNodes).ToList();
+        Assert.AreEqual(1, allRefs.Count());
```
enikao, line 52: What about using `CollectionsAssert`?

test/LionWeb-CSharp-Test/tests/ReferenceExtensionsTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        var refs = ReferenceExtensions.FindIncomingReferences(circle, [bom]).ToList();
+        Assert.AreEqual(1, refs.Count());
+        var ref0 = refs.First();
+
+        var expectedReferenceValue =
+            new ReferenceValue(bom, language.BillOfMaterials_materials, 0, circle);
+        Assert.AreEqual(expectedReferenceValue, ref0);  // (relies on value equality of C# record types)
+    }
+}
```
enikao, line 80: I'd add a tests:
* we find a reference to itself
* we find two references in the same parent in different references
* we find two references in the same multi-reference (yes, we can have multiple times the same target)
* Defined behavior if we give the same target twice
* Defined behavior if we give the same scope node twice
* It works with both (targets.size and scope.size) > 2 [yay! COBOL expressions]

test/LionWeb-CSharp-Test/tests/ReferenceExtensionsTests.cs
dslmeinte: Didn't know about that one, but will use it.

test/LionWeb-CSharp-Test/tests/ReferenceExtensionsTests.cs
enikao: Do you want to add these in this PR? Otherwise looks good.

dslmeinte: Added them

dslmeinte:
> One more proposed test: Not using Dynamic nodes, so we're confident we don't rely on any implementation specific to them.

You mean: not rely on `DynamicNode` at all? In that case, I'd make a new, simple test language.

enikao:
> > One more proposed test: Not using Dynamic nodes, so we're confident we don't rely on any implementation specific to them.
> 
> You mean: not rely on `DynamicNode` at all? In that case, I'd make a new, simple test language.

Yes. You could also use any of the references from ShapesLanguage, whatever suits you more.

enikao (approved):

PR #19, opened 2024-07-09
Remove List<INode?> constructor and give Comparere more "immutable" API

enikao:


test/LionWeb-CSharp-Test/tests/SerializationTests.cs
```diff
@@ -51,6 +51,27 @@ public void test_serialization_shapes_model()
...
+
+    [TestMethod]
+    public void test_serialization_lioncore()
+    {
+        var serializationChunk = LanguageSerializer.Serialize(M3Language.Instance);
+        Console.WriteLine(JsonUtils.WriteJsonToString(serializationChunk));
+
+        // Just run the deserializer for now (without really checking anything), to see whether it crashes or not:
+        LanguageDeserializer.Deserialize(serializationChunk);
+    }
```
dslmeinte, line 73: Could also run the comparer here, right?

test/LionWeb-CSharp-Test/tests/SerializationTests.cs
```diff
@@ -51,6 +51,27 @@ public void test_serialization_shapes_model()
         new Deserializer([_language]).Deserialize(serializationChunk);
     }
 
+    [TestMethod]
+    public void test_serialization_shapes_language()
+    {
+        var serializationChunk = LanguageSerializer.Serialize(ShapesLanguage.Instance);
+        Console.WriteLine(JsonUtils.WriteJsonToString(serializationChunk));
```
dslmeinte, line 58: Can you use an `ITestOutputHelper` or something like that here?

dslmeinte (approved):

test/LionWeb-CSharp-Test/tests/SerializationTests.cs
enikao: For what? I'm missing the context.

dslmeinte: To write to that instead of to the console. Or does this testing framework faithfully capture the console output per test method?

test/LionWeb-CSharp-Test/tests/SerializationTests.cs
enikao: Nope, doesn't work with M3

test/LionWeb-CSharp-Test/tests/SerializationTests.cs
enikao: seems to work for me

Issue #20, opened 2024-07-10
Clean up generation of test languages


dslmeinte:
Current gripes:

- Several (forms of) definitions of the Shapes language floating around
- `LionWeb-CSharp-Generator-Test.Generate.cs` is not run as part of anything, so generated sources are outdated. (The generated sources could also be generated into `LionWeb-CSharp-Test`.)
- PowerShell _and_ shell scripts that do the same thing => only use PowerShell.
- No documentation.

Also: create a central `make.ps1` that triggers all generation.

dslmeinte:
Fixed by PR #22

PR #22, opened 2024-07-10
Clean up generation

dslmeinte:
See issue #20


README.md
```diff
@@ -11,12 +11,13 @@ This repository contains an implementation of (select parts of) the [LionWeb](ht
 
 * [`src/LionWeb-CSharp`]: implementation of LionWeb for/in C#
 * [`src/LionWeb-CSharp-Generator`]: implementation of a C# code generator for LionWeb
-* [`build/LionWeb-CSharp-build`]: a .NET console application that generates various source files in the `-Test` project
+* [`build/LionWeb-CSharp-Build`]: a .NET console application that generates various source files in the `-Test` project
```
enikao, line 14: These kind of changes virtually guarantee trouble on case-maintaining (but not case-sensitive) file systems, especially most windows machines. I see the inconsistency, but I'm not sure fixing them is worth the trouble for all users.

build/LionWeb-CSharp-Build/README.md
```diff
@@ -0,0 +1,12 @@
+# README
```
enikao, line 1: I don't really get why it's better to generate into another project, than that project generating into itself.

enikao:
Do we want to have an _actual_ test for the generator, i.e. comparing the generated result?

README.md
dslmeinte: True, but currently we don't have many so virtually everyone actually cloning the repo (and not just ingesting the NuGet packages) will only ever see the renamed folder.

build/LionWeb-CSharp-Build/README.md
dslmeinte: Because if there are compilation errors in the generated code, the project will fail to build, and then the generator will not run. (You could Git-revert the changes due to generation to start over again.) Plus: I don't like state/self-modifying stuff in general.

dslmeinte:
> Do we want to have an _actual_ test for the generator, i.e. comparing the generated result?

I don't really feel the need, as we already have that: the generated code for the test languages is checked in, so any change to it is visible as a Git diff.

make.ps1
```diff
@@ -0,0 +1,7 @@
+cd build/LionWeb-CSharp-Build
+pwsh build.ps1
```
enikao, line 2: When I try this:

```
PS lionweb-csharp> .\make.ps1
pwsh : The term 'pwsh' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At lionweb-csharp\make.ps1:2 char:1
+ pwsh build.ps1
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (pwsh:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

make.ps1
enikao, line 1: Seems not to be part of the solution (can't see it in Rider)

build/LionWeb-CSharp-Build/build.ps1
```diff
@@ -4,19 +4,15 @@ if (-Not (Test-Path $LocalChunks)) {
...
 
-dotnet run Build.cs --no-dependencies --no-restore #--watch
+dotnet run Generate.cs --no-dependencies --no-restore
 
 $ExtDefs = "chunks/externalDefs";
 $LocalDefs = "chunks/localDefs";
 $TestProjectDefs = $TestProject + "/languages/defChunks";
 
-cp $LocalDefs/shapesLanguage.json $TestProjectDefs/shapes.json
+mv $LocalDefs/shapes.json $TestProjectDefs/
```
enikao, line 17: When executing:

```
mv : Cannot create a file when that file already exists.
At lionweb-csharp\build\LionWeb-CSharp-build\build.ps1:17 char:1
+ mv $LocalDefs/shapes.json $TestProjectDefs/
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : WriteError: (...efs\shapes.json:FileInfo) [Move-Item], IOException
    + FullyQualifiedErrorId : MoveFileInfoItemIOError,Microsoft.PowerShell.Commands.MoveItemCommand
```

make.ps1
dslmeinte: Ah, right — works outside of Windows. Can you check whether changing that line to `./build.ps1` works for you?

enikao: yes, works

build/LionWeb-CSharp-Build/build.ps1
dslmeinte: Different semantics of POSIX-compliant `mv` (on macOS) vs. `mv` on Windows. I'll replace this with PowerShell-builtin commands.

make.ps1
dslmeinte: That's weird, because it _is_ checked in: https://github.com/LionWeb-io/lionweb-csharp/tree/feature/clean-up-generation/test/LionWeb-CSharp-Test/languages/defChunks

enikao: VS solutions have their own interpretation of the file system -- I pushed the addition.

enikao (approved):

Issue #21, opened 2024-07-10
Query methods ancestor, preceding-sibling, following-sibling


erk-dik:
Hi @enikao, @dslmeinte, @JojoEffect,  

As far as I am concerned, the current implementation lacks the following methods, as explained below:

Ancestor: queries the ancestor nodes filtered by a type, something like:
```
public static T? Ancestor<T>(this INode node, bool includeSelf = false) where T : INode =>
    node.Ancestors(includeSelf)
        .OfType<T>()
        .FirstOrDefault();
```
PrecedingSibling  (PrevSiblings) / FollowingSibling  (NextSiblings): queries previous/next siblings of a node, with an optional parameter that specifies whether to include the current node.

enikao:
The question is whether we should implement this in lionweb-csharp or a downstream project. Do you think it's generally useful?

enikao:
Compare to XPath axes: https://www.w3.org/TR/1999/REC-xpath-19991116/#axes

PR #23, opened 2024-07-12
Add ancestor, precedingSibling and followingSibling M1 extension methods

erk-dik:
This PR includes the following M1 extension methods (mentioned in issue #21) and their tests.

- Ancestor
- PrecedingSiblings
- FollowingSiblings

### DIFF
--- a/src/LionWeb-CSharp/core/M1/M1Extensions.cs
+++ b/src/LionWeb-CSharp/core/M1/M1Extensions.cs
@@ -88,43 +88,41 @@ public static class M1Extensions
         parent.Set(containment, list);
     }
     
-    //TODO: The order of returned siblings ? 0 ... n or n ...0
+    /// <summary>
+    /// Returns all the preceding siblings of the base node <param name="self"></param>
+    /// Optionally includes <paramref name="self"/>.
+    /// 
+    /// Starting from index 0 to the index of base node, includes the index of <param name="self"></param>
+    /// if <paramref name="includeSelf"/> is set to true.  
+    /// </summary>
+    /// <param name="self">Base node to find preceding siblings of.</param>
+    /// <param name="includeSelf">If true, the result includes <paramref name="self"/>.</param>
+    /// <returns>List of preceding siblings of <param name="self"></param>.</returns>
+    /// <exception cref="TreeShapeException">If <paramref name="self"/> has no parent, or is not a child in a multiple containment.</exception>
     public static IEnumerable<INode> PrecedingSibling(this INode self, bool includeSelf = false)
     {
-        INode? parent = self.GetParent();
-        if (parent == null)
-            throw new TreeShapeException(self, "Cannot get siblings of a node with no parent");
-        
-        Containment containment = parent.GetContainmentOf(self)!;
-        if (!containment.Multiple)
-            throw new TreeShapeException(self, "A single containment does not have siblings");
-        
-        var value = parent.Get(containment);
-        if (value is not IEnumerable enumerable)
-            // should not happen
-            throw new TreeShapeException(self, "A single containment does not have siblings");
-        
-        var list = new List<INode>(containment.AsNodes<INode>(enumerable));
-        return includeSelf ? list.Take(list.IndexOf(self) + 1) : list.Take(list.IndexOf(self));
+        List<INode> list = GetContainmentNodes(self);
+        var takeCount = includeSelf ? list.IndexOf(self) + 1 : list.IndexOf(self);
+        return list.Take(takeCount);
     }
-    
+
+    /// <summary>
+    /// Returns all the following siblings of the base node <param name="self"></param>
+    /// Optionally includes <paramref name="self"/>.
+    ///
+    /// Starting from the following index of <param name="self"></param>, includes the index of <paramref name="self"/>
+    /// if <paramref name="includeSelf"/> is set to true.  
+    /// </summary>
+    ///
+    /// <param name="self">Base node to find following siblings of.</param>
+    /// <param name="includeSelf">If true, the result includes <paramref name="self"/>.</param>
+    /// <returns>List of following siblings of <param name="self"></param>.</returns>
+    /// <exception cref="TreeShapeException">If <paramref name="self"/> has no parent, or is not a child in a multiple containment.</exception>
     public static IEnumerable<INode> FollowingSibling(this INode self, bool includeSelf = false)
     {
-        INode? parent = self.GetParent();
-        if (parent == null)
-            throw new TreeShapeException(self, "Cannot get siblings of a node with no parent");
-        
-        Containment containment = parent.GetContainmentOf(self)!;
-        if (!containment.Multiple)
-            throw new TreeShapeException(self, "A single containment does not have siblings");
-        
-        var value = parent.Get(containment);
-        if (value is not IEnumerable enumerable)
-            // should not happen
-            throw new TreeShapeException(self, "A single containment does not have siblings");
-        
-        var list = new List<INode>(containment.AsNodes<INode>(enumerable));
-        return includeSelf ? list.Skip(list.IndexOf(self)) : list.Skip(list.IndexOf(self) + 1);
+        List<INode> list = GetContainmentNodes(self);
+        var skipCount = includeSelf ? list.IndexOf(self) : list.IndexOf(self) + 1;
+        return list.Skip(skipCount);
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -87,6 +87,45 @@ public static void InsertAfter(this INode self, INode newSuccessor)
         list.Insert(index + 1, newSuccessor);
         parent.Set(containment, list);
     }
+    
+    //TODO: The order of returned siblings ? 0 ... n or n ...0
+    public static IEnumerable<INode> PrecedingSibling(this INode self, bool includeSelf = false)
+    {
+        INode? parent = self.GetParent();
```
enikao, line 94: We now have the same code in 4 methods, definitely time to factor it out.

src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -87,6 +87,45 @@ public static void InsertAfter(this INode self, INode newSuccessor)
...
+        if (!containment.Multiple)
+            throw new TreeShapeException(self, "A single containment does not have siblings");
+        
+        var value = parent.Get(containment);
+        if (value is not IEnumerable enumerable)
+            // should not happen
+            throw new TreeShapeException(self, "A single containment does not have siblings");
+        
+        var list = new List<INode>(containment.AsNodes<INode>(enumerable));
+        return includeSelf ? list.Take(list.IndexOf(self) + 1) : list.Take(list.IndexOf(self));
```
enikao, line 108: `includeSelf` only affects the parameter to `Take()`, so move the condition there.

src/LionWeb-CSharp/core/M1/M1Extensions.cs
erk-dik: Repeated code factored out forPrecedingSiblings and FollowingSibliblings, but not for InsertAfter and InsertBefore, the reasons are the following:

-  Factored out method return a list of nodes to forPrecedingSiblings and FollowingSibliblings, but InsertAfter and InsertBefore makes use of parent, containment variables after the non-repeated code block.
- exception messages are different for PrecedingSiblings, FollowingSibliblings and  InsertAfter and InsertBefore. Of course those could be parameters to the factored out method.

test/LionWeb-CSharp-Test/tests/nodeapi/AxisTests.cs
```diff
@@ -138,6 +138,164 @@ public void ManyParents_Annotation_Self()
...
+    
+    [TestMethod]
+    public void Ancestor_ManyParents()
+    {
+        var node = new Coord("n");
+        var parent = new Line("p") { Start = node };
+        var ancestor = new Geometry("a") { Shapes = [parent] };
+        Assert.IsNotNull(node.Ancestor<Shape>());
+        Assert.AreEqual(parent, node.Ancestor<Shape>());
+    }
```
JojoEffect, line 167: Should the `ancestor` (`Geometry`) be tested here too?

test/LionWeb-CSharp-Test/tests/nodeapi/AxisTests.cs
```diff
@@ -138,6 +138,164 @@ public void ManyParents_Annotation_Self()
...
+    
+    [TestMethod]
+    public void Ancestor_ManyParents_Self()
+    {
+        var node = new Coord("n");
+        var parent = new Line("p") { Start = node };
+        var ancestor = new Geometry("a") { Shapes = [parent] };
+        Assert.IsNotNull(node.Ancestor<Shape>(true));
+        Assert.AreEqual(parent, node.Ancestor<Shape>(true));
+    }
```
JojoEffect, line 177: Should the `ancestor` (`Geometry`) be tested here too?

src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -164,6 +215,24 @@ public static void InsertAfter(this INode self, INode newSuccessor)
     public static IEnumerable<INode> Ancestors(this INode self, bool includeSelf = false) =>
         Ancestors<INode>(self, includeSelf);
```
JojoEffect, line 217: @enikao lets have a short talk about this exceptions being thrown here if the given node does not fulfill all reuirements. I'm not sure if a user expects this (even if it is documented at the API)

test/LionWeb-CSharp-Test/tests/nodeapi/AxisTests.cs
JojoEffect, line 1: Is this filename a good one? Maybe I'm just not used to the term "Axis" in the context of models/trees/graphs!?

erk-dik: Niko would better comment on this, but to my knowledge the term "axis" comes from XML path language (XPath) specs : https://www.w3.org/TR/1999/REC-xpath-19991116/#axes

enikao: Erkan quotes the right source; we could use the correct plural "Axes".

test/LionWeb-CSharp-Test/tests/nodeapi/AxisTests.cs
erk-dik: Added.

test/LionWeb-CSharp-Test/tests/nodeapi/AxisTests.cs
erk-dik: @enikao shall I rename the test file ? ( AxisTests.cs -> AxesTests.cs )

JojoEffect: Suggestion from @enikao and me: "TreeTraversalTests" or "TTT" or "T³" (last both are just a joke)

JojoEffect (approved):

src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -87,6 +87,43 @@ public static void InsertAfter(this INode self, INode newSuccessor)
...
+    /// Optionally includes <paramref name="self"/>.
+    /// 
+    /// Starting from index 0 to the index of base node, includes the index of <param name="self"></param>
+    /// if <paramref name="includeSelf"/> is set to true.  
+    /// </summary>
+    /// <param name="self">Base node to find preceding siblings of.</param>
+    /// <param name="includeSelf">If true, the result includes <paramref name="self"/>.</param>
+    /// <returns>List of preceding siblings of <param name="self"></param>.</returns>
+    /// <exception cref="TreeShapeException">If <paramref name="self"/> has no parent, or is not a child in a multiple containment.</exception>
+    public static IEnumerable<INode> PrecedingSibling(this INode self, bool includeSelf = false)
```
enikao, line 102: We return more than one, so name should be plural: `PrecedingSiblings`

src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -87,6 +87,43 @@ public static void InsertAfter(this INode self, INode newSuccessor)
...
+    ///
+    /// Starting from the following index of <param name="self"></param>, includes the index of <paramref name="self"/>
+    /// if <paramref name="includeSelf"/> is set to true.  
+    /// </summary>
+    ///
+    /// <param name="self">Base node to find following siblings of.</param>
+    /// <param name="includeSelf">If true, the result includes <paramref name="self"/>.</param>
+    /// <returns>List of following siblings of <param name="self"></param>.</returns>
+    /// <exception cref="TreeShapeException">If <paramref name="self"/> has no parent, or is not a child in a multiple containment.</exception>
+    public static IEnumerable<INode> FollowingSibling(this INode self, bool includeSelf = false)
```
enikao, line 121: plural

src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -154,6 +191,20 @@ public static void InsertAfter(this INode self, INode newSuccessor)
         bool includeAnnotations = false) =>
         Children<INode>(self, includeSelf, includeAnnotations);
 
+    
+    /// <summary>
+    /// Enumerates all direct and indirect parents of <paramref name="self"/> that matches with type <typeparamref name="T"/>
```
enikao, line 196: Single return, so no enumeration

src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -154,6 +191,20 @@ public static void InsertAfter(this INode self, INode newSuccessor)
...
+    /// Optionally includes <paramref name="self"/>.
+    /// </summary>
+    /// <param name="self">Base node to find ancestor of.</param>
+    /// <param name="includeSelf">If true, the result includes <paramref name="self"/>.</param>
+    /// <typeparam name="T"></typeparam>
+    /// <returns>The first ancestor of <paramref name="self"/> that matches with type <typeparamref name="T"/> or default </returns>
```
enikao, line 202: use `<c>null</c>` instead of "default", way easier to understand (and always correct)

enikao (approved):

Issue #29, opened 2024-10-07
Generate `@Obsolete` attributes


dslmeinte:
As soon as `@Deprecated` is exported (see https://github.com/LionWeb-io/lionweb-mps/issues/104), we can put corresponding `@Obsolete` attributes in the generated C# code.

enikao:
implemented in #30

PR #30, opened 2024-10-14
generate @Deprecated annotation -> [Obsolete] attribute

enikao:
Implements #29


test/LionWeb-CSharp-Test/languages/generated/ALang.g.cs
```diff
@@ -25,23 +25,23 @@ public ALangLanguage(string id) : base(id)
 		_aEnum_right = new(() => new EnumerationLiteralBase<ALangLanguage>("id-right", AEnum, this) { Key = "key-right", Name = "right" });
 	}
 
-	/// <inheritdoc/>
+	/// <inheritdoc/>
```
dslmeinte, line 28: I'm a little mistified by these ∂s: there's no change to the generator basics/utils, but somehow there's a diff that's not showing. Is it Windows vs. Unix newlines? Are you doing this on Windows on macOS (where I'm seeing the same)?

It's not a biggie, but it's confusing to have such a (trivial) diff after re-generation.

dslmeinte (approved):

test/LionWeb-CSharp-Test/languages/generated/Deprecated.g.cs
```diff
@@ -0,0 +1,350 @@
...
+
+	/// <inheritdoc/>
+        public override Enum GetEnumerationLiteral(EnumerationLiteral literal)
+	{
+		if (_language.DeprEnum.EqualsIdentity(literal.GetEnumeration()))
+			return EnumValueFor<DeprEnum>(literal);
+		throw new UnsupportedEnumerationLiteralException(literal);
+	}
+
+	public virtual DeprConcept NewDeprConcept(string id) => new(id);
```
JojoEffect, line 126: Idea: could you also mark the respective factory method as obsolete if the corresponding concept is deprecated? Or are ther some issues with that approach?

test/LionWeb-CSharp-Test/languages/generated/Deprecated.g.cs
```diff
@@ -0,0 +1,350 @@
...
+{
+	private string? _deprProp = null;
+	/// <remarks>Optional Property</remarks>
+        [LionCoreMetaPointer(Language = typeof(DeprecatedLanguage), Key = "MDkzNjAxODQtODU5OC00NGU3LTliZjUtZmIxY2U0NWE0ODBhLzc4MTUyNDM0Nzk0ODc5OTM0NDYvNzgxNTI0MzQ3OTQ4Nzk5MzQ0OQ")]
+	[LionCoreFeature(Kind = LionCoreFeatureKind.Property, Optional = true, Multiple = false)]
+	[Obsolete("deprProp comment")]
+	public string? DeprProp { get => _deprProp; set => SetDeprProp(value); }
+
+	/// <remarks>Optional Property</remarks>
+        public DeprConcept SetDeprProp(string? value)
```
JojoEffect, line 146: Same question as about the factory methods here again for all the generated methods?

JojoEffect (commented):
I like it! :-) thanks! Only the question remains if it is possible to mark quite "all" occurences  of a deprecated "thing" - e.g. the generated methods for th "thing", too?

test/LionWeb-CSharp-Test/languages/generated/ALang.g.cs
enikao: I don't see this diff in Rider, and I'm not curious enough to dig out a hex editor.

enikao:
Good improvement @JojoEffect! I hope I covered all places now?

JojoEffect (approved):
Nice - looks good! Thanks :-)

Issue #31, opened 2024-10-16
Make all generated public interfaces, classes and structs *partial*


JojoEffect:
To be able to more easily extend the generated concepts and interfaces and additionally to easily add the non-generatable parts of the languages (e.g. the behaviours) it would be helpful to have the generated types represented as `public partial class|interface|struct XYZ` types.

JojoEffect:
@enikao I would expect this to be a small extension to the generator - Do you think it is possible to have this implemented in the very near future?

JojoEffect:
> Do we generate `struct`s anywhere?

Not as far as I know - just wanted make the requirement more general. I'm fully satisfied if non-existing structs are not made partial 😬

enikao:
Implemented in #32

PR #32, opened 2024-10-16
generate classifiers as partial C# types

enikao:
implements #31

### DIFF
--- a/test/LionWeb-CSharp-Test/languages/generated/ALang.g.cs
+++ b/test/LionWeb-CSharp-Test/languages/generated/ALang.g.cs
@@ -99,7 +99,7 @@ public class ALangFactory : AbstractBaseNodeFactory, IALangFactory
 /// crip
 ///    tion</summary>
 [LionCoreMetaPointer(Language = typeof(ALangLanguage), Key = "key-AConcept")]
-public class AConcept : NodeBase
+public partial class AConcept : NodeBase
 {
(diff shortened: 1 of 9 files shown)


dslmeinte (approved):

test/LionWeb-CSharp-Test/languages/generated/ALang.g.cs
JojoEffect, line 1: It would be nice to have the same ability on the factories, so one can either extend the factories by encapsulation + inheritance or just by adding/overriding content to the factory. In our concrete use case could potentially reduce the numer of additiona interfaces and abstractions. What do you think?

JojoEffect, line 1: I'm not sure if it would also help on the generated `*Language` type. What do you think? Or will this open some doors we probably want to keep closed?

enikao: I don't see an issue with that, although also not that much of an advantage -- we can extend the factory and set the new one as default on the language.

enikao: There is not much internal to the `Language` class that I could imagine exploitable. However, it kind of invites adding methods there because it's convenient, not because they belong there.

JojoEffect (approved):

PR #33, opened 2024-10-31
hardened node api axis against circular containment

enikao:


src/LionWeb-CSharp/core/M1/M1Extensions.cs
```diff
@@ -175,9 +175,19 @@ public static T ReplaceWith<T>(this INode self, T replacement) where T : INode
     /// <param name="includeSelf">If true, the result includes <paramref name="self"/>.</param>
     /// <param name="includeAnnotations">If true, the result includes directly and indirectly contained annotations.</param>
     /// <returns>All directly and indirectly contained nodes of <paramref name="self"/>.</returns>
+    /// <exception cref="TreeShapeException">If containment hierarchy contains cycles.</exception>
     public static IEnumerable<INode> Descendants(this INode self, bool includeSelf = false,
         bool includeAnnotations = false) =>
-        Descendants<INode>(self, includeSelf, includeAnnotations);
+        new ResettingEnumerable(self, includeSelf, includeAnnotations);
+
+    private class ResettingEnumerable(INode self, bool includeSelf, bool includeAnnotations) : IEnumerable<INode>
```
dslmeinte, line 183: Why is this class necessary to have? Is it an optimization that effectively memoises the result of the `Descendants` call?

dslmeinte (approved):

src/LionWeb-CSharp/core/M1/M1Extensions.cs
enikao: Otherwise, if somebody reuses the result (double enumeration) , they get false positive `TreeShapeException`s.
Test at `CircularContainmentTests.NonCircularMultipleEnumerationDescendants()`

dslmeinte: Might be good to spend a line of documentation on that.

enikao: also added as XDoc

PR #36, opened 2024-11-18
Language extensions to create serialization chunk

erk-dik:
This is a mini-DSL to simplify the creation of serialization chunks

### DIFF
--- a/src/LionWeb-CSharp/core/serialization/Serialization.cs
+++ b/src/LionWeb-CSharp/core/serialization/Serialization.cs
@@ -25,7 +25,7 @@ public record SerializationChunk
 {
     public required string SerializationFormatVersion { get; init; }
     public required SerializedLanguageReference[] Languages { get; init; }
-    public SerializedNode[] Nodes { get; init; }
+    public SerializedNode[] Nodes { get; init; } = [];
 }
 
 public record SerializedLanguageReference
@@ -38,10 +38,10 @@ public record SerializedNode
 {
     public required string Id { get; init; }
     public required MetaPointer Classifier { get; init; }
-    public SerializedProperty[] Properties { get; init; }
-    public SerializedContainment[] Containments { get; init; }
-    public SerializedReference[] References { get; init; }
-    public string[] Annotations { get; init; }
+    public SerializedProperty[] Properties { get; init; } = [];
+    public SerializedContainment[] Containments { get; init; } = [];
+    public SerializedReference[] References { get; init; } = [];
+    public string[] Annotations { get; init; } = [];
     public string? Parent { get; init; }
 }
 


src/LionWeb-CSharp/core/serialization/Serialization.cs
```diff
@@ -25,7 +25,7 @@ public record SerializationChunk
 {
     public required string SerializationFormatVersion { get; init; }
     public required SerializedLanguageReference[] Languages { get; init; }
-    public required SerializedNode[] Nodes { get; init; }
+    public required SerializedNode[] Nodes { get; set; }
```
enikao, line 28: We don't want to change production code for tests.

enikao:
Maybe we've missed the most obvious solution: Set only the actually required properties to `required`, e.g. node id/classifier.
That doesn't change the API or semantics, as there are sensible defaults for the others.

src/LionWeb-CSharp/core/serialization/Serialization.cs
```diff
@@ -37,10 +38,10 @@ public record SerializedNode
 {
     public required string Id { get; init; }
     public required MetaPointer Classifier { get; init; }
-    public required SerializedProperty[] Properties { get; init; }
-    public required SerializedContainment[] Containments { get; init; }
-    public required SerializedReference[] References { get; init; }
-    public required string[] Annotations { get; init; }
+    public SerializedProperty[] Properties { get; init; }
```
enikao, line 41: You need to initialize these, e.g. `public SerializedProperty[] Properties { get; init; } = [];`

test/LionWeb-CSharp-Test/languages/SerializationChunkExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+// 
+// SPDX-FileCopyrightText: 2024 TRUMPF Laser SE and other contributors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace LionWeb_CSharp_Test.languages;
+
+using LionWeb.Core.M3;
+using LionWeb.Core.Serialization;
+
+public static class SerializationChunkExtensions
```
enikao, line 23: If you go this route, you need to go full Builder.

So you'd need a `SerializationChunkBuilder` class, with a method `SerializedNodeBuilder Node(string id)`.

Usage would be like:

```
var serializedChunk = SerializationChunkBuilder.Create()
  .NewNode("myId")  // returns SerializedNodeBuilder
    .MetaPointer("l", "v", "k")
    .Parent(null)
    .NewProperty()  // returns SerializedPropertyBuilder
      .Property(myProp)
      .Value(1)
    .NewProperty() // this is cheating, as it's on SerializedPropertyBuilder, but acts on its owning SerializedNodeBuilder
      .Property("l","v","k")
      .Value("myVal")
    .NewContainment() // again cheating; returns SerializedContainmentBuilder
      .Containment(myCont)
      .Targets("aId", "bId", "cId")
      .Target(myNode)
  .build();
```

test/LionWeb-CSharp-Test/tests/serialization/SerializationChunkExtensionsTests.cs
```diff
@@ -0,0 +1,279 @@
...
+
+*/
+
+[TestClass]
+public class SerializationChunkExtensionsTests
+{
+    [TestMethod]
+    public void test_nodes_with_Concept()
+    {
+        var serializationChunk = new SerializationChunk
```
enikao, line 40: This doesn't need any builders, and is already way easier to read.
Maybe that's already sufficient, even without builders.

You could think about a few convenience methods, like `SerializedLanguageReference Serialize(this Language lang)`, then your example becomes

```
var lang = ShapesLanguage.Instance;
var serializationChunk = new SerializationChunk
{
  SerializationFormatVersion = ReleaseVersion.Current,
  Languages = [ lang.Serialize() ],
  Nodes =
  [
    new SerializedNode
    {
      Id = "A",
      Classifier = lang.Coord.Serialize()
    },
    new SerializedNode
    {
      Id = "B",
      Classifier = lang.Circle.Serialize()
    }
  ]
};
```

enikao:
@erk-dik Shall we give this PR up?

erk-dik:
> @erk-dik Shall we give this PR up?

I think so, I was having the same thought.

PR #35, opened 2024-11-16
Release 2024.1 : StructuredDatatype

enikao:

PR #37, opened 2024-11-19
Support multipe versions of languages and LionWeb

enikao:
See readme.md for changes


enikao:
Merging, otherwise continuing #35 is too cumbersome.

PR #38, opened 2024-12-02
Generator improvements

enikao:
* added `Language.SetFactory()`
* sort all generated language entries
* generate factory and language as partial class
* improved LanguageDeserializer

### DIFF
--- a/build/LionWeb-CSharp-Build/Generate.cs
+++ b/build/LionWeb-CSharp-Build/Generate.cs
@@ -51,6 +51,7 @@ var testLanguagesDefinitions = new TestLanguagesDefinitions();
 var aLang = testLanguagesDefinitions.ALang;
 var bLang = testLanguagesDefinitions.BLang;
 var tinyRefLang = testLanguagesDefinitions.TinyRefLang;
+var mpsSpecific = DeserializeExternalLanguage("Io.Lionweb.Mps.Specific").First();
 
 List<Names> names =
 [
@@ -73,6 +74,7 @@ List<Names> names =
     },
     new (tinyRefLang, "Examples.TinyRefLang"),
     new (deprecatedLang, "Examples.DeprecatedLang"),
+    // new (mpsSpecific, "Io.Lionweb.Mps.Specific")
 ];
 
 
--- a/src/LionWeb-CSharp-Generator/impl/OrderedExtensions.cs
+++ b/src/LionWeb-CSharp-Generator/impl/OrderedExtensions.cs
@@ -22,18 +22,20 @@ using Core.M3;
 
 public static class OrderedExtensions
 {
+    private const string _separator = ".";
+
     public static IEnumerable<Feature> Ordered(this IEnumerable<Feature> features) =>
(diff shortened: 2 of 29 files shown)


build/LionWeb-CSharp-Build/Generate.cs
```diff
@@ -73,6 +74,7 @@ DynamicLanguage[] DeserializeExternalLanguage(string name, params Language[] dep
     },
     new (tinyRefLang, "Examples.TinyRefLang"),
     new (deprecatedLang, "Examples.DeprecatedLang"),
```
dslmeinte, line 76: (trailing comma)

src/LionWeb-CSharp-Generator/impl/OrderedExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace LionWeb.CSharp.Generator.Impl;
+
+using Core.M2;
+using Core.M3;
+
+public static class OrderedExtensions
+{
+    public static IEnumerable<Feature> Ordered(this IEnumerable<Feature> features) =>
+        features.OrderBy(f => f.GetLanguage().Name + f.GetFeatureClassifier().Name + f.Name);
```
dslmeinte, line 26: Maybe concatenate these fragments but separated with a character that's sorted before any legal character of a name string?

test/LionWeb-CSharp-Test/tests/ModelAccessTests.cs
```diff
@@ -101,8 +101,8 @@ public void test_AllDescendants()
         Assert.IsFalse(allDescendants.Contains(geometry));
         var line = geometry.Shapes[0] as Line;
         Assert.AreEqual(line, allDescendants.ElementAt(0));
-        Assert.AreEqual(line.Start, allDescendants.ElementAt(1));
-        Assert.AreEqual(line.End, allDescendants.ElementAt(2));
+        Assert.AreEqual(line.End, allDescendants.ElementAt(1));
```
dslmeinte, line 104: Doesn't this mean comparisons like these are inherently brittle? Can we come up with assertions that don't rely on the order of the elements in `allDescendants`?

dslmeinte (approved):
(some minor, cosmetic comments)

test/LionWeb-CSharp-Test/tests/ModelAccessTests.cs
enikao: AFAIK, we don't state anything about the order of decendants. I used the less brittle variant in the other tests -- `CollectionAssert.AreEquivalent()` instead of `CollectionAssert.AreEqual()`. We could use the same here.

The order changed because of the alphabetical sorting of features.

build/LionWeb-CSharp-Build/Generate.cs
enikao: ... which is very useful if we want to re-activate `mps.specific`

src/LionWeb-CSharp-Generator/impl/OrderedExtensions.cs
enikao: fixed

PR #47, opened 2024-12-25
added special interfaces for instances of Annotations and Partitions

enikao:
After our decision in https://github.com/LionWeb-io/specification/issues/320, annotation instances are structurally different from other nodes -- leading to `IAnnotationInstance`.

In https://github.com/LionWeb-io/specification/issues/269 we've decided that delta subscriptions are based on partitions, thus partition instances get a special role. The new interface `IPartitionInstance` prepares for that.


src/LionWeb.Core/Core/BaseTypes.cs
```diff
@@ -78,6 +78,32 @@ public interface IReadableNode<out T> : IReadableNode where T : IReadableNode
     public new IReadOnlyList<T> GetAnnotations();
 }
 
+/// Instance of an <see cref="Annotation"/>.
+/// <inheritdoc />
+public interface IAnnotationInstance : IReadableNode
```
dslmeinte, line 83: Don't we need `IConceptInstance`, if only for the sake of symmetry?

dslmeinte (approved):

PR #48, opened 2025-01-06
Add event listeners to API

enikao:
Basic idea: We have listeners to forests of partitions (`IForestListener`) and to partitions (`IPartitionListener`).
They use C# standard events.

Forest stuff is not implemented yet.

For any change, a node tries to find the partition it is contained in: by walking up the containment tree until it finds the first instance of `IPartitionInstance`. If the partition has a `IPartitionCommander`, the change is sent via the commander.
This currently works for changes to properties, references, and containments, both via generated methods and via reflection (`Set()`).

For the latter, we need to detect changes between the already existing feature values, and the newly set ones. We do this in several steps:
1. `ListComparer` uses [Hirschberg's algorithm](https://en.wikipedia.org/wiki/Hirschberg%27s_algorithm) to find the smallest set of changes;
2. `StepwiseListComparer` converts the total list indices into list indices per step; 
3. `SetContainmentEvent` / `SetReferenceEvent` convert the changes per step into commands.

For some commands, we need to record some information _before_ the actual change to the nodes, and only emit the events _after_ the change. We want to keep event logic separated from model update logic. Thus, we use specializations of `PartitionEventBase` for such cases. In all change operations, we first check the preconditions for the change. Then we call `PartitionEventBase.CollectOldData()`, and afterwards execute the node change. Finally, we call `PartitionEventBase.RaiseEvent()`.

`PartitionEventHandler` implements both `IPartitionInstance` and `IPartitionCommander`. It forwards each command to the corresponding event.

Some of the pre-processing of event data is pretty expensive (especially the list comparison), so we want to skip it if nobody would receive it. I didn't find a native way to ask C# whether there are any subscribers to an event. Thus, `IPartitionCommander` also offers `CanRaiseXxx()` methods; they use my custom `CountingEventHandler` to keep track of the number of subscribers.
We call the appropriate variant of these methods any time we check for an available `IPartitionCommander`.


src/LionWeb.Core/Core/M1/EventHandler.cs
```diff
@@ -0,0 +1,602 @@
...
+    public event EventHandler<IPartitionListener.PropertyAddedArgs> PropertyAdded
+    {
+        add => _propertyAdded.Add(value);
+        remove => _propertyAdded.Remove(value);
+    }
+
+    private readonly CountingEventHandler<IPartitionListener.PropertyAddedArgs> _propertyAdded = new();
+
+    /// <inheritdoc />
+    public void AddProperty(IWritableNode node, Property property, PropertyValue newValue) =>
```
dslmeinte, line 109: Would it be useful to have a formal description of the deltas (&rArr; events, commands) so that we can generate types for them in the various languages?

dslmeinte:
You do list comparison for when `Set(...)` is used to manipulate a multi-valued link feature directly (instead of through dedicated (, generated) methods which already know exactly what events/∂s to spew out), right?

src/LionWeb.Core/Core/M1/EventHandler.cs
dslmeinte: Not sure whether it's not worth it. Might try something in a time-boxed way, because I'm hitting the same problem in the TS code.

enikao:
> You do list comparison for when `Set(...)` is used to manipulate a multi-valued link feature directly (instead of through dedicated (, generated) methods which already know exactly what events/∂s to spew out), right?

Yes, only `SetContainmentEvent` and `SetReferenceEvent` use list comparison. Other implementations like `SingleContainmentEvent`, `AddMultipleContainmentsEvent`, and `NodeBase.RaiseReferenceAddEvent()` leverage the additional information they have.

enikao:
I separated the fixes for list comparison into #49. This _could_ be merged, but I might make a new release before that.

enikao:
superseded by #95

PR #53, opened 2025-01-16
introduced optionally uncompressed ids during deserialization

enikao:


src/LionWeb.Core/Core/M1/CompressedId.cs
```diff
@@ -26,17 +26,104 @@ namespace LionWeb.Core.M1;
...
-    public bool IsIdDuplicate(CompressedId compressedId) =>
+    public bool IsIdDuplicate(ICompressedId compressedId) =>
         !_knownIds.Add(compressedId);
 }
 
+/// <summary>
+/// A potentially compressed id.
+/// Implementation differ in their memory vs. computation tradeoffs.
+/// </summary>
+public interface ICompressedId
```
dslmeinte, line 40: The name seems slightly misleading: an instance of `ICompressedId` might not be compressed at all. How about `IPersistedId`? Or even `[I]Id`? And `IdConfig` instead of `CompressedIdConfig`?

test/LionWeb.Core.Test/Deserialization/CustomHandler/InvalidFeatureTests.cs
```diff
@@ -199,7 +199,7 @@ public void invalid_feature_heals_to_a_containment()
         IDeserializer deserializer = new DeserializerBuilder()
             .WithHandler(deserializerHealingHandler)
             .WithLanguage(ShapesLanguage.Instance)
-            .WithUncompressedIds(true)
+            .WithCompressedIds(new(KeepOriginal:true))
```
dslmeinte, line 202: [COSMETIC] Whitespace missing; should be: `new(KeepOriginal: true)`. There's multiple occurrences of this further on as well.

src/LionWeb.Core/Core/M1/CompressedId.cs
```diff
@@ -26,17 +26,104 @@ namespace LionWeb.Core.M1;
...
+/// <summary>
+/// An uncompressed id that always stores its original.
+///
+/// <para>
+/// More memory efficient for models with typical ids longer than 20 characters,
+/// but a bit faster.
+/// </para>
+/// </summary>
+/// <param name="original">Original, uncompressed id.</param>
+public readonly struct UncompressedId(string original) : ICompressedId, IEquatable<UncompressedId>
```
dslmeinte, line 86: `OriginalId` as name might also be an option?

test/LionWeb.Core.Test/Serialization/StreamingTests.cs
```diff
@@ -34,8 +34,8 @@ public StreamingTests()
         _language = ShapesLanguage.Instance;
     }
 
-    // private const long _maxSize = 1_500_000L;
-    private const long _maxSize = 1_500L;
+    private const long _maxSize = 1_500_000L;
+    // private const long _maxSize = 1_500L;
```
dslmeinte: [COSMETIC] (commented-out code)

test/LionWeb.Core.Test/Serialization/StreamingTests.cs
```diff
@@ -97,9 +97,11 @@ public void MassDeserialization()
         // Assembly.GetExecutingAssembly().GetManifestResourceStream("LionWeb.Core.Test.resources.expected.json") ??
```
dslmeinte, line 97: [COSMETIC] (commented-out code)

test/LionWeb.Core.Test/Serialization/StreamingTests.cs
```diff
@@ -97,9 +97,11 @@ public void MassDeserialization()
...
         //     throw new AssertFailedException();
 
-        var deserializer = new DeserializerBuilder()
-            .WithLanguage(_language)
-            .Build();
+        var deserializer = new Deserializer(LionWebVersions.v2024_1) { CompressedIdConfig = new(KeepOriginal: true) };
+            // .WithLanguage(_language)
```
dslmeinte, line 101: [COSMETIC] (commented-out code)

test/LionWeb.Core.Test/Serialization/UncompressedIdTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+using M1;
+
+[TestClass]
+public class UncompressedIdTests
+{
+    [TestMethod]
+    public void Equals()
+    {
+        var left = ICompressedId.Create("a", new CompressedIdConfig(false, false));
```
dslmeinte, line 28: `new CompressedIdConfig(*, *)` is not super-readable. Could add static instances to `CompressedIdConfig` such as `CompressedWithOriginal`, `CompressedWithoutOriginal`, etc.

src/LionWeb.Core/Core/M1/Deserializer.cs
```diff
@@ -42,8 +42,8 @@ public partial class Deserializer : DeserializerBase<IWritableNode, IDeserialize
     private readonly Dictionary<ICompressedId, List<ICompressedId>> _annotationsByOwnerId = new();
 
     /// <inheritdoc />
-    public Deserializer(LionWebVersions lionWebVersion, IDeserializerHandler? handler = null) :
-        base(lionWebVersion, handler ?? new DeserializerExceptionHandler())
+    public Deserializer(LionWebVersions lionWebVersion, IDeserializerHandler? handler = null, CompressedIdConfig? compressedIdConfig = null) :
```
dslmeinte, line 45: Why not `= new()` instead of `= null`? Or is the creation of the default argument value not lazy?

src/LionWeb.Core/Core/M1/CompressedId.cs
```diff
@@ -26,17 +26,104 @@ namespace LionWeb.Core.M1;
...
+    public static ICompressedId Create(string id, bool keepOriginal) =>
+        Create(id, new CompressedIdConfig(KeepOriginal: keepOriginal));
+}
+
+/// <summary>
+/// Configuration which optimizations to apply to (potentially compressed) ids.
+/// </summary>
+/// <param name="Compress">Whether we compress ids at all; defaults to <c>false</c>.</param>
+/// <param name="KeepOriginal">Whether we keep the original around for compressed ids; defaults to <c>false</c>. Uses more memory, but eases debugging.</param>
+public record CompressedIdConfig(bool Compress = false, bool KeepOriginal = false);
```
dslmeinte, line 75: `CompressedIdConfig(false, B)` has only one sensible value for `B`: `true`.

dslmeinte (approved):
Approved, but with questions.

src/LionWeb.Core/Core/M1/CompressedId.cs
enikao: The proper name would be `IPotentiallyCompressedId`, which is quite a mouthful. I'd like to keep "compressed" in there somehow, because that's the _reason d'etre_ for this interface.

test/LionWeb.Core.Test/Serialization/StreamingTests.cs
enikao: yes, will stay like this: To meaningfully test we need a LOT of nodes, but that's annoying when running all tests all the time.

src/LionWeb.Core/Core/M1/CompressedId.cs
dslmeinte: lolcatspeak-compliant suggestion: `IAreCompressed[?]Id` 😸

test/LionWeb.Core.Test/Serialization/UncompressedIdTests.cs
enikao: if we want this readable use named parameters, no evil constants.

src/LionWeb.Core/Core/M1/Deserializer.cs
enikao: not allowed by c#

src/LionWeb.Core/Core/M1/CompressedId.cs
enikao: Correct. So this should be documented?

PR #56, opened 2025-01-20
Fix language deserializer annotations

enikao:


build/LionWeb-CSharp-Build/Generate.cs
```diff
@@ -43,6 +43,7 @@
     var withEnumLanguage = DeserializeExternalLanguage(lionWebVersion, "with-enum").First();
     withEnumLanguage.Name = "WithEnum";
     var specificLanguage = ISpecificLanguage.Get(lionWebVersion);
+    // var specificLanguage = DeserializeExternalLanguage(lionWebVersion, "MpsSpecific-metamodel-annotated").First();
```
dslmeinte, line 46: Can you comment why the commented-out code would be worth keeping around? (Otherwise, I'd delete it.)

src/LionWeb.Generator/VersionSpecific/V2023_1/io.lionweb.mps.specific.V2023_1.g.cs
```diff
@@ -6,30 +6,30 @@
...
+using LionWeb.Core.Utilities;
+using LionWeb.Core.VersionSpecific.V2023_1;
 using System;
 using System.Collections.Generic;
 
 [LionCoreLanguage(Key = "io-lionweb-mps-specific", Version = "0")]
 public partial class SpecificLanguage : LanguageBase<ISpecificFactory>, ISpecificLanguage
 {
-	public static readonly SpecificLanguage Instance = new Lazy<SpecificLanguage>(() => new("io-lionweb-mps-specific")).Value;
+	internal static readonly SpecificLanguage Instance = new Lazy<SpecificLanguage>(() => new("io-lionweb-mps-specific")).Value;
```
dslmeinte, line 21: I can't see whether the `io.lionweb.mps.specific` language actually changed. I'm guessing all these diffs are the result of generating with the most recent generator, and the two metamodel versions are just the result of making an export in either version. Is that correct? In which case: shouldn't there be a changed MPS file somewhere?

dslmeinte (approved):
with questions

src/LionWeb.Generator/VersionSpecific/V2023_1/io.lionweb.mps.specific.V2023_1.g.cs
enikao: Mostly re-generation. The only real change is that the 2024.1 variant now uses language version `2024.1`.

PR #58, opened 2025-02-03
Implemented generic subtree hasher

enikao:
should work as a base for a cross-implementation spec

### DIFF
--- a/src/LionWeb.Core/Core/Utilities/Hasher.cs
+++ b/src/LionWeb.Core/Core/Utilities/Hasher.cs
@@ -222,8 +222,12 @@ public class Hasher
     #endregion
 }
 
+/// A small piece of data representing the state of an ordered set of nodes.
+/// Any change to the underlying nodes results in a different hash.
+/// <seealso cref="Hasher"/>.
 public interface IHash : IEquatable<IHash>;
 
+/// <inheritdoc cref="IHash" />
 public readonly record struct ByteArrayHash(byte[] Hash) : IHash
 {
     /// <inheritdoc />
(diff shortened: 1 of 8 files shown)


JojoEffect (approved):
Looks good. Thanks!

src/LionWeb.Core/Core/Utilities/Hasher.cs
```diff
@@ -0,0 +1,240 @@
...
+///   <item>feature: meta-pointer, value</item>
+///   <item>property: serialized value</item>
+///   <item>single containment: same node</item>
+///   <item>multiple containment: same list of nodes</item>
+///   <item>single reference with internal target: internal index</item> 
+///   <item>single reference with external target: node id</item> 
+///   <item>multiple reference: same reference at each position</item>
+/// </list>
+/// </summary>
+public class Hasher
```
dslmeinte, line 52: This class does _two_ things: visiting every node in a sub tree, and updating a hash according to the visited nodes. Would it be possible to separate those two things?

dslmeinte (approved):
Looks good: left one comment.

src/LionWeb.Core/Core/Utilities/Hasher.cs
enikao: I think we had a similar discussion for the comparer: For me, the iteration is so trivial that it doesn't warrant a special class. Also, the differences in _how_ to traverse, and what data to keep how during traversal, is too varied to make a useful abstraction.

dslmeinte: OK, although I think triviality largely lives in the eyes of the beholder 😄

PR #61, opened 2025-02-07
Generate `bool TryGetMyFeature(out MyFeatureType? myFeature)` methods

enikao:
for _multiple_ features, the signature is `bool TryGetMyFeature out IReadOnlyList<MyFeatureType> myFeature)`.


test/LionWeb.Core.Test/NodeApi/Generated/ReferenceTests_Multiple_Optional.cs
```diff
@@ -280,6 +289,14 @@ public void Null_Remove_Empty()
         Assert.ThrowsException<InvalidValueException>(() => parent.RemoveShapes(null));
     }
 
+    [TestMethod]
+    public void Null_TryGet()
```
dslmeinte, line 293: Wouldn't `None_TryGet` be a more accurate name for this test method? (And likewise for the other `Null_TryGet`s on multi-valued links.)

enikao: yes, but all other methods in, and the section itself, is called `null`

PR #64, opened 2025-03-12
generate KeyedDescription annotations to Xdoc

enikao:


src/LionWeb.Generator/Impl/AstExtensions.cs
```diff
@@ -358,6 +356,33 @@ public static IEnumerable<XmlNodeSyntax> XdocLine(string text, string? tag = nul
         ];
     }
 
+    /// <returns><c>///&lt;seealso cref="target"/&gt;</c></returns>
+    public static IEnumerable<XmlNodeSyntax> XdocSeeAlso(TypeSyntax target)
+    {
+        var body = XmlSeeAlsoElement(NameMemberCref(target));
```
dslmeinte, line 362: (Could be inlined, and then the whole method could use `=>` syntax.)

src/LionWeb.Generator/Impl/AstExtensions.cs
```diff
@@ -358,6 +356,33 @@ public static IEnumerable<XmlNodeSyntax> XdocLine(string text, string? tag = nul
...
+            XdocSlashes(),
+            body,
+            XdocNewline()
+        ];
+    }
+
+    /// <returns><c>///&lt;seealso href="uri"/&gt;</c></returns>
+    public static IEnumerable<XmlNodeSyntax> XdocSeeAlso(string uri)
+    {
+        var body = XmlEmptyElement("seealso")
```
dslmeinte, line 375: (Could be inlined, and then the whole method could use `=>` syntax.)

PR #67, opened 2025-04-17
made M1Extensions.ReplaceWith() work with annotation instances

enikao:


dslmeinte (approved):
AFAIC, the changelog addition can be in the same commit as the one where the work has been done.

JojoEffect (commented):
👍

PR #69, opened 2025-04-17
introduced generic TryGet(Feature) API

enikao:


test/LionWeb.Core.Test/M2ReflectionTests.cs
```diff
@@ -231,8 +231,8 @@ public void Language_Set_Entities()
     public void Language_TryGet_Entities()
     {
         var node = new DynamicLanguage("a", _lionWebVersion);
-        Assert.IsTrue(((Language)node).TryGetEntities(out var empty));
```
dslmeinte, line 234: Can't that cast (and all other, similar ones) not be moved to line 233?

src/LionWeb.Core/Core/M3/Types.cs
```diff
@@ -120,11 +120,11 @@ public interface Classifier : LanguageEntity
...
     /// </summary>
     /// <param name="features">Value of <see cref="Features"/> if set, or <c>null</c>.</param>
-    /// <returns><c>true</c> if <see cref="Features"/> is set; <c>false</c> otherwise.</returns>
-    bool TryGetFeatures(out IReadOnlyList<Feature>? features)
+    /// <returns><c>true</c> if <see cref="Features"/> is set and not empty; <c>false</c> otherwise.</returns>
+    bool TryGetFeatures([NotNullWhen(true)] out IReadOnlyList<Feature>? features)
     {
         features = Features;
-        return true;
+        return features is { Count: > 0 };
```
dslmeinte, line 127: There's probably a reason that the semantics have been changed to include “and not empty” here, but it eludes me: could you clarify?

dslmeinte (changes requested):

src/LionWeb.Core/Core/M3/Types.cs
enikao: It's in line with the behavior of generated code: If a multiple link is required and empty, the getter throws.

test/LionWeb.Core.Test/M2ReflectionTests.cs
enikao: No, otherwise this line doesn't work:

```
node.AddEntities([new DynamicConcept("b", _lionWebVersion, null)]);
```

PR #71, opened 2025-04-17
added LionWebVersions 2025.1 and 2025.1-compatible

enikao:
As of now, content is same as for 2024.1 / 2024.1-compatible

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres _loosely_ to [Semantic Versioning](https://semver.org/s
 ## [Unreleased]
 
 ### Added
+* Added `LionWebVersions` _2025.1_ and _2025.1-compatible_. No content differences to _2024.1_ so far.
 ### Fixed
 ### Changed
 ### Removed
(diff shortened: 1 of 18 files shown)


dslmeinte (approved):
I don't enjoy the code duplication, though! I get it why it's an obvious choice, but I'd much rather would have that there are some base classes that capture the commonality — which is currently 100% between 2024.1 and 2025.1, anyway — and so also clearly state the difference.

enikao:
> I don't enjoy the code duplication, though!

Yes, I agree. IMHO, there's no really good solution here -- either code duplication, or even more convoluted indirection. As the whole "externalize this logic from its actual place to version-specifics" already introduces a lot of indirection, I opted for cutting it here.

Another reason: Old versions should _never_ change their behavior (outside of bug fixes), so it should be a very conscious decision to update old versions.

dslmeinte:
> > I don't enjoy the code duplication, though!
> 
> Yes, I agree. IMHO, there's no really good solution here -- either code duplication, or even more convoluted indirection. As the whole "externalize this logic from its actual place to version-specifics" already introduces a lot of indirection, I opted for cutting it here.

I was thinking the same as disadvantage of DRY'ing the usual way.

> Another reason: Old versions should _never_ change their behavior (outside of bug fixes), so it should be a very conscious decision to update old versions.

Not that having the code duplicated, but still not protected against modification gives a solid guarantee.

PR #73, opened 2025-04-17
enable replacing DynamicLanguage.entities; improve DynamicStructuredDataType

enikao:


src/LionWeb.Core/Core/M3/DynamicTypes.cs
```diff
@@ -1110,6 +1129,35 @@ public void AddDependsOn(IEnumerable<Language> languages)
...
+
+    /// <inheritdoc />
+    protected override bool DetachChild(INode child)
+    {
+        if (base.DetachChild(child))
+        {
+            return true;
+        }
+
+        var c = GetContainmentOf(child);
```
dslmeinte, line 1141: (Cosmetic: why not "`containment`" instead of "`c`"?)

PR #74, opened 2025-04-17
hardened LenientNode against changing keys

enikao:
C# Dictionary seems to cache the result of GetHashCode() in some weird ways.

### DIFF
--- a/src/LionWeb.Core/Core/LenientNode.cs
+++ b/src/LionWeb.Core/Core/LenientNode.cs
@@ -20,6 +20,7 @@ namespace LionWeb.Core;
 using M2;
 using M3;
 using System.Collections;
+using System.Diagnostics.CodeAnalysis;
 using Utilities;
 
 /// <summary>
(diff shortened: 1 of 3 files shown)


src/LionWeb.Core/Core/LenientNode.cs
```diff
@@ -295,4 +321,44 @@ public override void InsertAnnotations(int index, IEnumerable<INode> annotations
     /// <inheritdoc />
     public override bool RemoveAnnotations(IEnumerable<INode> annotations) =>
         RemoveSelfParent(annotations?.ToList(), _annotations, null);
+
+    private IEnumerable<Feature> FeatureKeys => _featureValues.Select(f => f.feature);
+
+    private bool TryGetFeature(Feature featureToFind, out object? value)
```
dslmeinte, line 327: I find the names of these methods slightly misleading: `TryGetFeature**Value**` etc. would be better — as in: clearer.

dslmeinte (approved):
(but see remark)

src/LionWeb.Core/Core/LenientNode.cs
enikao: I'll leave it, as it's `private`.

PR #75, opened 2025-04-17
provide API to access serializationFormatVersion during deserialization

enikao:


src/LionWeb.Core/Core/Serialization/JsonUtils.cs
```diff
@@ -86,7 +89,15 @@ public static async Task<List<IReadableNode>> ReadNodesFromStreamAsync(Stream st
                     await Advance();
                     string? version = streamReader.GetString();
                     if (version != null)
-                        deserializer.LionWebVersion.AssureCompatible(version);
+                    {
```
dslmeinte, line 92: OCD-thing, but I find this `if` statement much easier to read with a `== null` guard, and that also matches the documentation exactly.

src/LionWeb.Core/Core/Serialization/JsonUtils.cs
```diff
@@ -72,8 +72,11 @@ public static void WriteNodesToStream(Stream stream, ISerializer serializer, IEn
     /// </summary>
     /// <param name="stream">Stream to read from.</param>
     /// <param name="deserializer">Deserializer to use.</param>
+    /// <param name="lionWebVersionChecker">Optional action to check the <see cref="SerializationChunk.SerializationFormatVersion"/>.
```
dslmeinte, line 75: Maybe document that the given `Action` implementation should throw on an incompatibility?

dslmeinte (approved):
(but see remarks)

PR #77, opened 2025-04-17
Introduced SerializerBuilder and LanguageDeserializerBuilder

enikao:


build/LionWeb-CSharp-Build/Generate.cs
```diff
@@ -123,6 +123,9 @@ DynamicLanguage[] DeserializeExternalLanguage(LionWebVersions lionWebVersion, st
     SerializationChunk serializationChunk =
         JsonUtils.ReadJsonFromString<SerializationChunk>(
             File.ReadAllText($"chunks/externalDefs/{lionWebVersion.VersionString}/{name}.json"));
-    return new LanguageDeserializer(lionWebVersion, compressedIdConfig: new(KeepOriginal: true))
+    return new LanguageDeserializerBuilder()
```
dslmeinte, line 126: What's the advantage of using the Builder Pattern over using a configuration object (with reasonable effective defaults)?

enikao: That's a question for @JojoEffect -- seems to be more C#-ish.

enikao: I think one reason is that the Microsoft text editor can kinda handle auto-completion for methods, but mostly fails on anything more complex, like property initializers.

PR #78, opened 2025-04-24
Introduced ComparerBehaviorConfig.CompareCompatibleClassifier

enikao:
Introduced ComparerBehaviorConfig.CompareCompatibleClassifier to configure whether instances of C# Annotation and Concept should be considered different.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres _loosely_ to [Semantic Versioning](https://semver.org/s
 * Introduced `SerializerBuilder`.
 * Introduced `LanguageDeserializerBuilder`.
 * Introduced delegating implementations for `SerializerHandler`, `DeserializerHandler`, `LanguageDeserializerHandler`.
+* Introduced `ComparerBehaviorConfig.CompareCompatibleClassifier` to configure whether 
+  instances of C# `Annotation` and `Concept` should be considered different.
 ### Fixed
 * `LenientNode` now works properly if keys of features change.
 * Deserializer can now create instances of languages not registered beforehand.
--- a/test/LionWeb.Core.Test/Utilities/Comparer/ComparerNodeTests.cs
+++ b/test/LionWeb.Core.Test/Utilities/Comparer/ComparerNodeTests.cs
@@ -73,6 +73,24 @@ public class ComparerNodeTests : ComparerTestsBase
     }
 
     [TestMethod]
+    public void Annotation_Concept_Ignored()
+    {
+        var left = lF.NewDocumentation("b");
+        var right = rF.NewLine("a");
+
+        var parent = new NodeDifference(left, right);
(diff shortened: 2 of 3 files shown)


test/LionWeb.Core.Test/Utilities/Comparer/ComparerNodeTests.cs
```diff
@@ -72,6 +72,24 @@ public void Annotation_Concept()
...
     }
 
+    [TestMethod]
+    public void Annotation_Concept_Ignored()
+    {
+        var left = lF.NewDocumentation("b");
+        var right = rF.NewLine("a");
+
+        var parent = new NodeDifference(left, right);
+        IDifference[] differences = [parent, new IncompatibleClassifierDifference(left, right) { Parent = parent } ];
```
dslmeinte, line 82: `differences` isn't used anywhere, right?

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres _loosely_ to [Semantic Versioning](https://semver.org/s
 * Introduced `SerializerBuilder`.
 * Introduced `LanguageDeserializerBuilder`.
 * Introduced delegating implementations for `SerializerHandler`, `DeserializerHandler`, `LanguageDeserializerHandler`.
+* Introduced `ComparerBehaviorConfig.CompareCompatibleClassifier` to configure whether
```
dslmeinte, line 19: I find “compatible” a bit vague here: IMHO it doesn't capture the semantics succinctly. Can we find something better?

test/LionWeb.Core.Test/Utilities/Comparer/ComparerNodeTests.cs
```diff
@@ -72,6 +72,24 @@ public void Annotation_Concept()
...
+        Comparer comparer = new Comparer([(IReadableNode?)left], [right])
+        {
+            BehaviorConfig = new()
+            {
+                CompareCompatibleClassifier = false
+            }
+        };
+        Assert.IsTrue(comparer.AreEqual(), comparer.ToMessage(OutputConfig));
```
dslmeinte, line 90: But _structurally_ `left` and `right` differ as well, right? So, why does the comparer deem them to be equal?

dslmeinte (changes requested):

CHANGELOG.md
enikao: I had the same issue already with `IncompatibleClassifierDifference`, didn't find anything better then.

dslmeinte: At least it's consistent(ly badly-named), then :)

test/LionWeb.Core.Test/Utilities/Comparer/ComparerNodeTests.cs
enikao: You're right, good spot! Fixed it.

dslmeinte (approved):

PR #79, opened 2025-04-24
ensure dynamically created C# enums have proper LionCoreMetaPointer attributes

enikao:


src/LionWeb.Core/Core/M2/Factories.cs
```diff
@@ -212,33 +212,44 @@ public override INode CreateNode(string id, Classifier classifier) =>
...
 
-        return CreateEnum(literal);
+        return CreateEnumLiteral(literal);
     }
 
-    private Enum CreateEnum(EnumerationLiteral literal)
+    private Enum CreateEnumLiteral(EnumerationLiteral literal) =>
+        (Enum)Enum.Parse(CreateEnum(literal.GetEnumeration()), literal.Name);
+
+    protected Type CreateEnum(Enumeration enm)
```
dslmeinte, line 227: Naming is...“sub-optimal” IMHO:
1. The method names give no hint as to that a dynamic C# enum is being created.
2. Abbreviations `enm`, `lit`, etc. are ugly and unhelpful. At least consider `enm` &rarr; `lionCoreEnumeration`.

dslmeinte (approved):
But see remark!

PR #100, opened 2025-07-11
Better comparison of multiple links changed via Set(Feature, value)

enikao:
Thanks to @Ulyana-F1re for figuring out the hard algorithmic parts.


Ulyana-F1re (approved):
I mostly looked into the `ListComparer` folder, as this is what I am familiar with.

PR #101, opened 2025-07-14
implemented GetAvailableIds query

enikao:


src/LionWeb.Protocol.Delta/Repository/LionWebRepositoryBase.cs
```diff
@@ -28,6 +28,8 @@ public abstract class LionWebRepositoryBase<T>
     protected readonly Dictionary<string, IReadableNode> SharedNodeMap;
     protected readonly PartitionEventHandler PartitionEventHandler;
 
+    private long nextFreeNodeId = 12;
```
dslmeinte, line 31: Magic constant! Why not 0 (since anyway `repoProvidedId-${nextFreeNodeId++}`) or something large and round?

src/LionWeb.Protocol.Delta/Repository/LionWebRepositoryBase.cs
```diff
@@ -56,6 +58,20 @@ private void SendPartitionEventToAllClients(object? sender, IPartitionEvent? par
         SendAll(converted);
     }
 
+    protected IEnumerable<FreeId> GetFreeNodeIds(int count)
+    {
+        int returnedCount = 0;
+        while (returnedCount < count)
+        {
+            NodeId nextId = "repoProvidedId-" + nextFreeNodeId++;
```
dslmeinte, line 66: Unless synchronized, pre-incrementing is preferable/safer than post-incrementing.

dslmeinte (approved):
(but see remarks)

src/LionWeb.Protocol.Delta/Repository/LionWebRepositoryBase.cs
enikao: It's _a bit_ on purpose, so we don't expect 0 in tests and overlook some hard-coded implementation.
But then it should be in the `...Test` subclass, not in the production version.

PR #102, opened 2025-07-15
refactor insert containment logic

erk-dik:


src/LionWeb.Core/Core/M1/Event/Partition/PartitionEventReplicator.cs
```diff
@@ -326,25 +326,36 @@ private object ReplaceContainment(INode localParent, Containment containment, In
         return newValue;
     }
 
-    private object InsertContainment(INode localParent, Containment containment, Index index, INode nodeToInsert)
+    private static object InsertContainment(INode localParent, Containment containment, Index index, INode nodeToInsert)
     {
-        object newValue = nodeToInsert;
-        if (containment.Multiple)
+        object newValue;
```
dslmeinte, line 331: Can't you just `return <inlined: newValue expr>;` everywhere you do `newValue = <newValue expr>;` instead of the somewhat error-prone “late assignment and return of `newValue`”?

src/LionWeb.Core/Core/M1/Event/Partition/PartitionEventReplicator.cs
```diff
@@ -326,25 +326,36 @@ private object ReplaceContainment(INode localParent, Containment containment, In
...
-                {
-                    var children = new List<IWritableNode>(l.Cast<IWritableNode>());
-                    children.Insert(index, nodeToInsert);
-                    newValue = children;
-                }
-            } else
+            var existingChildren = localParent.Get(containment);
+            switch (existingChildren)
             {
-                newValue = new List<IWritableNode>() { nodeToInsert };
```
dslmeinte, line 345: If I read this correctly, this is not a refactoring but a functional change: if `existingChildren` is a single `IWriteableNode` then `newValue` will now be `nodeToInsert` instead of a list containing that value. Doesn't this affect unit tests, or don't those exist?

src/LionWeb.Core/Core/M1/Event/Partition/PartitionEventReplicator.cs
```diff
@@ -326,25 +326,36 @@ private object ReplaceContainment(INode localParent, Containment containment, In
...
+                    {
+                        var children = new List<IWritableNode>(l.Cast<IWritableNode>());
+                        children.Insert(index, nodeToInsert);
+                        newValue = children;
+                        break;
+                    }
+                case IWritableNode _:
+                    newValue = nodeToInsert;
+                    break;
+                default:
```
dslmeinte, line 348: Unit tests exercising all paths/situations would be good to have.

src/LionWeb.Core/Core/M1/Event/Partition/PartitionEventReplicator.cs
erk-dik: Actually, the previous code checks first whether the containment is multiple or single, and if not multiple, returns the nodeToInsert value, so code preserves that functionality .

src/LionWeb.Core/Core/M1/Event/Partition/PartitionEventReplicator.cs
erk-dik:
```
default:
    throw new InvalidValueException(containment, existingChildren);
```
How would you test this, a containment with an already assigned invalid value ? There must be sth going wrong between creation of the model and until it reaches here.

dslmeinte: In that case, adding a comment to that extent would be sufficient, but also necessary: defaults that never get reached (and so lack a unit test covering that code path) should be documented.

src/LionWeb.Core/Core/M1/Event/Partition/PartitionEventReplicator.cs
```diff
@@ -440,7 +442,7 @@ private void OnRemoteReferenceChanged(ReferenceChangedEvent referenceChangedEven
             localParent.Set(referenceChangedEvent.Reference, newValue);
         });
 
-    private object InsertReference(INode localParent, Reference reference, Index index, IReadableNode target)
+    private static object InsertReference(INode localParent, Reference reference, Index index, IReadableNode target)
```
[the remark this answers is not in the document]
erk-dik: Reasons provided by SonarQube were quite convincing: 

> Clarity and Intent: Marking a method/property as static makes it clear that the method does not depend on instance data and can be called without creating an instance of the class. This improves the readability of the code by clearly conveying the member’s intended use.
> Performance: Instance methods/properties in C# require an instance of the class to be called. This means that even if the it doesn’t use any instance data, the runtime still needs to pass a reference to the instance during the call. For static methods and properties, this overhead is avoided, leading to slightly better performance.
> Memory Usage: Since instance methods implicitly carry a reference to the instance (the caller object), they can potentially prevent the garbage collector from collecting the instance whem it is not otherwise referenced. Static members do not carry this overhead, potentially reducing memory usage.
> Testing: Static members can be easier to test since they do not require an instance of the class. This can simplify unit testing and reduce the amount of boilerplate code needed to set up tests.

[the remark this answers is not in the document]
erk-dik: In this case, they are tested implicitly via `public` methods. However, another approach would be: if they were `internal` instead of `private`, C# allows usage from a separate project (assembly) via `InternalsVisibleToAttribute `.

enikao (approved):

PR #104, opened 2025-07-21
colorful delta log output and better internal structures

enikao:


src/LionWeb.Protocol.Delta/Client/LionWebClientBase.cs
```diff
@@ -18,15 +18,29 @@
...
 
-public abstract class LionWebClientBase<T>
+public interface ILionWebClient
+{
+    private const string _magenta = "\x1b[95m";
+    private const string _bold = "\x1b[1m";
+    private const string _unbold = "\x1b[22m";
+    private const string _defaultColor = "\x1b[39m";
+    
+    public const string HeaderColor_Start = _magenta + _bold;
```
dslmeinte, line 34: Why not separate these things out, as they have overlap with `LionWebRepositoryBase`? See also e.g.  https://github.com/LionWeb-io/lionweb-typescript/blob/deltas/client-queue/packages/deltas-websocket/src/utils/ansi.ts

dslmeinte (approved):
(but see question!)

src/LionWeb.Protocol.Delta/Client/LionWebClientBase.cs
enikao: In theory, C# can do this natively: https://learn.microsoft.com/en-us/dotnet/api/system.console.foregroundcolor?view=net-9.0

In practice, it didn't work for me -- and I don't bother to battle that now.

PR #108, opened 2025-08-05
introducing ieventable interface

erk-dik:


src/LionWeb.Core/Core/BaseTypes.cs
```diff
@@ -209,31 +206,43 @@ public interface IWritableNode : IReadableNode
...
     /// <exception cref="InvalidValueException">If <paramref name="value"/> does not adhere to <paramref name="feature"/>'s type or constraints</exception>
     /// <seealso cref="IReadableNode.Get"/>
-    public void Set(Feature feature, object? value, IEventId? eventId = null);
+    public void Set(Feature feature, object? value);
 }
 
+/// <summary>
+/// Represents an interface for nodes that support event operations with optional event identifiers.
+/// </summary>
 public interface IEventableNode : IWritableNode
```
enikao, line 222: Let's move this to a separate file. This file should be taken apart anyways, so let's not add more to it.

src/LionWeb.Core/Core/BaseTypes.cs
```diff
@@ -209,31 +206,43 @@ public interface IWritableNode : IReadableNode
...
     /// Sets the given <paramref name="feature"/> on <c>this</c> node to the given <paramref name="value"/>.
     /// </summary>
     /// <exception cref="InvalidValueException">If <paramref name="value"/> does not adhere to <paramref name="feature"/>'s type or constraints</exception>
     /// <seealso cref="IReadableNode.Get"/>
-    public void Set(Feature feature, object? value, IEventId? eventId = null);
+    public void Set(Feature feature, object? value);
 }
 
+/// <summary>
+/// Represents an interface for nodes that support event operations with optional event identifiers.
```
enikao, line 220: what are "event operations"?

src/LionWeb.Core/Core/BaseTypes.cs
enikao: nitpicking: There are a few superfluous spaces in the new code.

src/LionWeb.Core/Core/BaseTypes.cs
```diff
@@ -422,6 +450,24 @@ public void DetachFromParent()
         }
     }
 
+    void IEventableNode.AddAnnotations(IEnumerable<IWritableNode> annotations, IEventId? eventId) =>
```
enikao, line 453: Why are these necessary?

enikao (commented):
We should have tests that cast a `NodeBase` instance to all of its super-interfaces, and call the respective variant of the method.

src/LionWeb.Core/Core/BaseTypes.cs
dslmeinte: As long as those namespace changes – **insofar these are really necessary**: multiple files can have the same namespace, right? – are in a separate commit, that's fine with me.

enikao: The namespace issue is the sole reason I didn't split the file yet. That would be a breaking change, so we'd need to tread carefully.

For now, let's only have the new stuff in a separate file.

test/LionWeb.Core.Test/NodeApi/NodeBaseTests.cs
```diff
@@ -0,0 +1,197 @@
...
+        
+        INode shape = new Circle("geom");
+        shape.AddAnnotations([doc, bom]);
+        
+        Assert.AreEqual(2, shape.GetAnnotations().Count);
+        CollectionAssert.AreEqual(new List<INode> { doc, bom }, shape.GetAnnotations().ToList());
+    }
+
+    [TestMethod]
+    public void AddAnnotations_IEventableNode()
```
enikao, line 54: We're missing `IWritableNode<>` and `IEventableNode<>`

src/LionWeb.Core/Core/BaseTypes.cs
enikao: Would we still need these if we _didn't_ implement `IWritable<>` inside `IEventable<>`?

src/LionWeb.Core/Core/IEventableNode.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace LionWeb.Core;
+
+using M1.Event;
+using M2;
+using M3;
+
+/// <summary>
+/// Represents an interface for nodes that support events with optional event identifiers.
+/// <seealso cref="IEvent"/>
```
enikao, line 26: should `seealso` be inside or outside `summary`?

src/LionWeb.Core/Core/BaseTypes.cs
enikao: With this, it seems to work without any additional methods in `NodeBase`:

```csharp
public interface IEventableNode<T> : IEventableNode, IWritableNode<T> where T : class, IEventableNode
{
    void IEventableNode.AddAnnotations(IEnumerable<IWritableNode> annotations, IEventId? eventId) => 
        AddAnnotations(M2Extensions.AsNodes<INode>(annotations), eventId);
    
    void IWritableNode.AddAnnotations(IEnumerable<IWritableNode> annotations) => 
        AddAnnotations(M2Extensions.AsNodes<INode>(annotations), null);
    
    void IWritableNode<T>.AddAnnotations(IEnumerable<T> annotations) => AddAnnotations(M2Extensions.AsNodes<T>(annotations));

    /// <inheritdoc cref="IWritableNode.AddAnnotations"/>
    /// <param name="eventId">The event ID of the event that triggers this action</param>
    public void AddAnnotations(IEnumerable<T> annotations, IEventId? eventId = null);

```

src/LionWeb.Core/Core/IEventableNode.cs
erk-dik: The right seems to be outside of summary.

test/LionWeb.Core.Test/Utilities/Hasher/HasherTests.cs
```diff
@@ -22,6 +22,7 @@ namespace LionWeb.Core.Test.Utilities.Hasher;
 using Languages.Generated.V2024_1.SDTLang;
 using Languages.Generated.V2024_1.Shapes.M2;
 using M1.Event;
+using M2;
```
enikao, line 25: necessary?

enikao (approved):

PR #109, opened 2025-08-11
rename delta types

erk-dik:


src/LionWeb.Core/Core/M1/Event/IDelta.cs
```diff
@@ -18,7 +18,7 @@
 namespace LionWeb.Core.M1.Event;
 
 /// Any event in the LionWeb event system.
-public interface IEvent
+public interface IDelta
```
dslmeinte, line 21: I don't get the rename — also no in relation to the `///` documentation.

Weren't all the `[I]<technical event name>Event` names for types (now being renamed to have a `Delta` suffix) directly taken from the specification? I tend to think of a “delta” as the thing that encapsulates all information about a model change, and the commands and events are CQRS-specific serializations (+ metadata, etc.) of that. So for me this renaming makes things more confusing.

PR #119, opened 2025-08-15
added Compositors

enikao:


src/LionWeb.Core/Core/Notification/CompositeNotification.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            builder.Append(part);
+        }
+        builder.Append(']');
+
+        return true;
+    }
+
+    public void Deconstruct(out IReadOnlyList<INotification> Parts, out INotificationId NotificationId)
```
dslmeinte, line 116: (Weird name for a method that performs something that I know as “destructuring”)

enikao: That's auto-generated by Rider, I presume it's a magical method in records.

dslmeinte: In that case: I miss a unicorn annotation 😁

PR #127, opened 2025-09-08
added benchmarking for node ids and (de)serializers

enikao:


erk-dik:
should benchmarking be part of  Lionweb.Core or standalone  project (e.g. LionWeb.Benchmarking) ?

test/LionWeb.Core.Benchmark/SerializerBenchmark.cs
```diff
@@ -0,0 +1,125 @@
...
+    [IterationSetup]
+    [TestInitialize]
+    public void CreateNodes()
+    {
+        _nodes = CreateNodes(_maxSize);
+    }
+
+    [Benchmark]
+    [TestMethod]
+    public async Task Serialize_Stream_Async()
```
erk-dik, line 43: This and Serialize_Stream have the same body except the last statements, can we extract them into a single method, or  it is not working in benchmarking ?

enikao: No, that's the `async` virus in action /-:

PR #136, opened 2025-09-24
Fix bug in ReplaceWith method and add replication support for various notifications/events

erk-dik:
Replicator now handles the following events:

- EntryMovedInSameReference
- AnnotationMovedAndReplacedInSameParent 
- AnnotationReplaced 
- ChildMovedAndReplacedInSameContainment

### DIFF
--- a/src/LionWeb.Core/Core/M1/M1Extensions.cs
+++ b/src/LionWeb.Core/Core/M1/M1Extensions.cs
@@ -141,6 +141,9 @@ public static class M1Extensions
         if (parent == null)
             throw new TreeShapeException(self, "Cannot replace a node with no parent");
 
+        if (replacement is null)
+            throw new ArgumentException("Null is not a valid node");
+        
         Containment containment = parent.GetContainmentOf(self);
 
         if (containment == null)
@@ -168,8 +171,19 @@ public static class M1Extensions
                 // should not happen
                 throw new TreeShapeException(self, "Node not contained in its parent");
 
-            nodes.Insert(index, replacement);
-            nodes.Remove(self);
+            var replacementParent = replacement.GetParent();
+            var replacementContainment = replacementParent?.GetContainmentOf(replacement);
+            if (replacementContainment == containment)
+            {
+                var replacementIndex = nodes.IndexOf(replacement);
+                nodes.Insert(index, replacement);
+                nodes.RemoveAt(index + 1);
+                nodes.RemoveAt(index < replacementIndex ? nodes.LastIndexOf(replacement) : nodes.IndexOf(replacement));
+            } else
+            {
+                nodes.Insert(index, replacement);
+                nodes.Remove(self);
+            }
             parent.Set(containment, nodes);
         } else
         {
--- a/src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
+++ b/src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
@@ -130,46 +130,46 @@ public class RemoteReplicator : NotificationPipeBase, INotificationHandler
 
     #region Partitions
 
-    private void OnRemoteNewPartition(PartitionAddedNotification partitionAdded) =>
-        SuppressNotificationForwarding(partitionAdded, () =>
+    private void OnRemoteNewPartition(PartitionAddedNotification notification) =>
+        SuppressNotificationForwarding(notification, () =>
         {
-            var newPartition = partitionAdded.NewPartition;
-            _localForest?.AddPartitions([newPartition], partitionAdded.NotificationId);
+            var newPartition = notification.NewPartition;
+            _localForest?.AddPartitions([newPartition], notification.NotificationId);
         });
 
-    private void OnRemotePartitionDeleted(PartitionDeletedNotification partitionDeleted) =>
-        SuppressNotificationForwarding(partitionDeleted, () =>
+    private void OnRemotePartitionDeleted(PartitionDeletedNotification notification) =>
+        SuppressNotificationForwarding(notification, () =>
         {
-            var localPartition = (IPartitionInstance?)LookupOpt(partitionDeleted.DeletedPartition.GetId());
+            var localPartition = (IPartitionInstance?)LookupOpt(notification.DeletedPartition.GetId());
             if (localPartition != null)
-                _localForest?.RemovePartitions([localPartition], partitionDeleted.NotificationId);
+                _localForest?.RemovePartitions([localPartition], notification.NotificationId);
         });
 
     #endregion
 
     #region Properties
 
-    private void OnRemotePropertyAdded(PropertyAddedNotification propertyAdded) =>
-        SuppressNotificationForwarding(propertyAdded, () =>
+    private void OnRemotePropertyAdded(PropertyAddedNotification notification) =>
+        SuppressNotificationForwarding(notification, () =>
         {
             Debug.WriteLine(
-                $"Node {propertyAdded.Node.PrintIdentity()}: Setting {propertyAdded.Property} to {propertyAdded.NewValue}");
-            Lookup(propertyAdded.Node.GetId()).Set(propertyAdded.Property, propertyAdded.NewValue,
-                propertyAdded.NotificationId);
+                $"Node {notification.Node.PrintIdentity()}: Setting {notification.Property} to {notification.NewValue}");
+            Lookup(notification.Node.GetId()).Set(notification.Property, notification.NewValue,
+                notification.NotificationId);
         });
 
-    private void OnRemotePropertyDeleted(PropertyDeletedNotification propertyDeleted) =>
-        SuppressNotificationForwarding(propertyDeleted, () =>
+    private void OnRemotePropertyDeleted(PropertyDeletedNotification notification) =>
+        SuppressNotificationForwarding(notification, () =>
         {
-            Lookup(propertyDeleted.Node.GetId())
-                .Set(propertyDeleted.Property, null, propertyDeleted.NotificationId);
+            Lookup(notification.Node.GetId())
+                .Set(notification.Property, null, notification.NotificationId);
         });
 
-    private void OnRemotePropertyChanged(PropertyChangedNotification propertyChanged) =>
-        SuppressNotificationForwarding(propertyChanged, () =>
+    private void OnRemotePropertyChanged(PropertyChangedNotification notification) =>
+        SuppressNotificationForwarding(notification, () =>
         {
-            Lookup(propertyChanged.Node.GetId()).Set(propertyChanged.Property, propertyChanged.NewValue,
-                propertyChanged.NotificationId);
+            Lookup(notification.Node.GetId()).Set(notification.Property, notification.NewValue,
+                notification.NotificationId);
         });
 
     #endregion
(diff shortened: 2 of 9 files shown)


src/LionWeb.Core/Core/M1/M1Extensions.cs
```diff
@@ -141,6 +141,9 @@ public static T ReplaceWith<T>(this INode self, T replacement) where T : INode
         if (parent == null)
             throw new TreeShapeException(self, "Cannot replace a node with no parent");
 
+        if (replacement is null)
+            throw new ArgumentException("Null is not a valid node");
```
enikao, line 145: We should use one of the Lionweb exceptions.

Also, the user doesn't know which node this message refers to.

src/LionWeb.Core/Core/M1/M1Extensions.cs
```diff
@@ -168,8 +171,19 @@ public static T ReplaceWith<T>(this INode self, T replacement) where T : INode
                 // should not happen
                 throw new TreeShapeException(self, "Node not contained in its parent");
 
-            nodes.Insert(index, replacement);
-            nodes.Remove(self);
+            var replacementParent = replacement.GetParent();
+            var replacementContainment = replacementParent?.GetContainmentOf(replacement);
+            if (replacementContainment == containment)
```
enikao, line 176: use `semanticEquals()` (or similar) extension method

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -114,181 +123,206 @@ public void Receive(INotificationSender correspondingSender, INotification notif
...
             switch (existingChildren)
             {
                 case IList l:
-                    var children = new List<IWritableNode>(l.Cast<IWritableNode>());
-                    children.Insert(index, substituteNode);
-                    children.RemoveAt(index + 1);
-                    return children;
-                case IWritableNode _ when index == 0:
+                    {
+                        var children = new List<IWritableNode>(l.Cast<IWritableNode>());
```
enikao, line 308: This feels like the same implementation in `M1Extensions.ReplaceWith()`. Can't we just use that?

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -335,94 +369,138 @@ private static object InsertContainment(INode localParent, Containment containme
...
         {
-            var localParent = Lookup(annotationMoved.Parent.GetId());
-            INode nodeToInsert = Lookup(annotationMoved.MovedAnnotation.GetId());
-            localParent.InsertAnnotations(annotationMoved.NewIndex, [nodeToInsert], annotationMoved.NotificationId);
+            var localParent = Lookup(notification.Parent.GetId());
+            INode nodeToInsert = Lookup(notification.MovedAnnotation.GetId());
+            localParent.InsertAnnotations(notification.NewIndex, [nodeToInsert], notification.NotificationId);
         });
 
+    private void OnAnnotationMovedAndReplacedInSameParent(AnnotationMovedAndReplacedInSameParentNotification notification) =>
```
enikao, line 412: Conceptually, this is the same as moveReplaceInSameContainment, but the implementation feels less involved. Is this implementation missing something or the other one overly complicated?

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -335,94 +369,138 @@ private static object InsertContainment(INode localParent, Containment containme
...
         });
+    
+    private void OnEntryMovedInSameReference(EntryMovedInSameReferenceNotification notification)
+    {
+        var localParent = Lookup(notification.Parent.GetId());
+        var target = Lookup(notification.Target.Reference.GetId());
+        
+        object newValue = target;
+        var reference = notification.Reference;
+        if (localParent.CollectAllSetFeatures().Contains(reference))
```
enikao, line 486: Can we use `TryGet(reference)`?

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -335,94 +369,138 @@ private static object InsertContainment(INode localParent, Containment containme
...
+        {
+            var existingTargets = localParent.Get(reference);
+            if (existingTargets is IList l)
+            {
+                var targets = new List<IReadableNode>(l.Cast<IReadableNode>());
+                targets.RemoveAt(notification.OldIndex);
```
enikao, line 492: This feels like it cannot handle all cases with old/new index being in front/behind the other

test/LionWeb.Core.Test/NodeApi/ReplaceTests.cs
```diff
@@ -17,12 +17,157 @@
...
+        };
+        
+        replaced.ReplaceWith(substituteNode);
+        
+        var expected = new Geometry("a")
+        {
+            Shapes = [new Line("substituteNode")]
+        };
+        
+        List<IDifference> differences = new Comparer(new List<INode?> {expected}, new List<INode?> {actual}).Compare().ToList();
```
enikao, line 48: let's introduce an `AssertEquals()` method for this repeated fragments

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -232,53 +238,124 @@ public void ChildReplaced_Deep()
...
     [TestMethod]
     public void ChildReplaced_Multiple_First()
     {
+        var replaced = new Circle("replaced");
+        var substituteNode = new Line("substituteNode");
+        
         var originalPartition = new Geometry("a")
         {
-            Shapes = [new Circle("replaced"), new Circle("child")]
+            Shapes = [replaced, new Circle("child")]
```
enikao, line 262: We can't distinguish the two circles

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -232,53 +238,124 @@ public void ChildReplaced_Deep()
...
 
+        Assert.AreEqual(1, notificationObserver.Count);
+        Assert.IsInstanceOfType<ChildReplacedNotification>(notificationObserver.Notifications[0]);
         AssertEquals([originalPartition], [clonedPartition]);
     }
 
     /// <summary>
-    /// This test triggers two notifications: ChildDeletedNotification, ChildAddedNotification
-    /// but it should trigger ChildReplacedNotification //TODO: requires fix !
+    /// Should trigger ChildReplacedNotification TODO: requires fix !
```
enikao, line 252: I'd prefer to implement the test as actually expected and ignore it

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -475,6 +584,94 @@ public void ChildMovedInSameContainment_Backward()
...
+    
+    [TestMethod]
+    public void ChildMovedAndReplacedInSameContainment_Backward_MoreThanThreeChildren()
+    {
+        var substituteNode = new Line("E");
+        var replaced = new Circle("B");
+        
+        var originalPartition = new Geometry("container")
+        {
+            Shapes = [new Circle("A"), replaced, new Circle("C"), new Circle("D"), substituteNode, new Circle("F")]
```
enikao, line 633: can't distinguish circles

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -475,6 +584,94 @@ public void ChildMovedInSameContainment_Backward()
...
+    
+    [TestMethod]
+    public void ChildMovedAndReplacedInSameContainment_Forward_MoreThanThreeChildren()
+    {
+        var substituteNode = new Line("E");
+        var replaced = new Circle("B");
+        
+        var originalPartition = new Geometry("container")
+        {
+            Shapes = [new Circle("A"), substituteNode, new Circle("C"), new Circle("D"), replaced, new Circle("F")]
```
enikao, line 658: can't distinguish circles

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -648,6 +845,116 @@ public void AnnotationMovedInSameParent_Backward()
         AssertEquals([originalPartition], [clonedPartition]);
     }
 
+    #endregion
+
+    #region AnnotationReplaced
+    
+    [TestMethod]
+    public void AnnotationReplaced_Multiple_Only()
```
enikao, line 853: This class is way too long, let's split it up

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -648,6 +845,116 @@ public void AnnotationMovedInSameParent_Backward()
...
+    {
+        var replaced = new BillOfMaterials("replaced");
```
enikao, line 855: we need to give the BOMs some feature to make them distinguishable

test/LionWeb.Core.Test/Notification/NotificationsTest.cs
```diff
@@ -648,6 +845,116 @@ public void AnnotationMovedInSameParent_Backward()
         AssertEquals([originalPartition], [clonedPartition]);
     }
 
+    #endregion
+
+    #region AnnotationReplaced
```
enikao, line 850: we should have the same set of test scenarios for multiple containments and annotations

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
erk-dik: Except some differences (ReplaceWith also handles the annotation) but it applies similar logic, part of the code that replaces node in the same containment is actually the same.

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
erk-dik: Thats true, but differences are:

- InsertAnnotations moves the node, different than list Insert, which requires removing the moved node from its old index.
- ReplaceContainment additionally checks singularity/plurality, and whether replaced and moved nodes are in the same containment or not. We do not need to check those for annotations.

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
erk-dik: we can not really use  ReplaceWith, unless we do some refactoring.

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
erk-dik: There is no replacement happening, here, maybe you have replacement case in your mind and thats why it feels like that :)

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
erk-dik: as agreed, this will be handled in another pr @enikao

dslmeinte (approved):
Looks entirely plausible, but mind you: I didn't go through these 4K lines with a comb.

I have two remarks on the CHANGELOG.

PR #137, opened 2025-09-25
Prevent redundant operations when references remain unchanged

erk-dik:
Added early return checks to avoid unnecessary notifications and operations when references or nodes are not modified.


src/LionWeb.Core/Core/M1/M1Extensions.cs
```diff
@@ -137,6 +137,9 @@ public static IEnumerable<INode> FollowingSiblings(this INode self, bool include
     /// <exception cref="TreeShapeException">If <paramref name="self"/> has no parent.</exception>
     public static T ReplaceWith<T>(this INode self, T replacement) where T : INode
     {
+        if (ReferenceEquals(self, replacement))
+            return (T)self;
```
enikao, line 141: You could return replacement without casting

enikao (approved):
Please add an entry to changelog

PR #140, opened 2025-09-30
revisit remote replicator (part 1)

erk-dik:
This PR revisits remote replicator and tests replicaton support for:

-  9 Child actions 
-  7 Annotation actions 
-  3 Property actions
-  2 Partition actions

Additionally tests files are reorganised, few noops test are added.

Actions marked in the image below: composite, change classifier and reference actions will be revisited in the next PR.
image

### DIFF
--- a/src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
+++ b/src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
@@ -398,12 +398,9 @@ public class RemoteReplicator : NotificationPipeBase, INotificationHandler
     private void OnRemoteAnnotationReplaced(AnnotationReplacedNotification notification) =>
         SuppressNotificationForwarding(notification, () =>
         {
-            //TODO: replace with replaceWith method
-            var localParent = Lookup(notification.Parent.GetId());
-            var newAnnotation = (INode)notification.NewAnnotation;
-            //var replacedAnnotation = (INode)notification.ReplacedAnnotation;
-            localParent.GetAnnotations()[notification.Index].ReplaceWith(newAnnotation);
-            //replacedAnnotation.ReplaceWith(newAnnotation);
+            var newAnnotation = LookupOpt(notification.NewAnnotation.GetId()) ?? (INode)notification.NewAnnotation;
+            var replacedAnnotation = Lookup(notification.ReplacedAnnotation.GetId());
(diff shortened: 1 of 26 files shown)


test/LionWeb.Core.Test/Notification/NoOperations/NoOpsTests_Property.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace LionWeb.Core.Test.Notification.NoOperationsTests;
+
+using Languages.Generated.V2025_1.Shapes.M2;
+
+[TestClass]
+public class NoOpsTests_Property: NotificationTestsBase
+{
+    [TestMethod]
+    public void PropertyAdded_with_null_value()
```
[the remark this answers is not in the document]
erk-dik: I added equality check resulting in noops, and added test.

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -216,20 +218,18 @@ private void OnRemoteChildDeleted(ChildDeletedNotification notification) =>
     private void OnRemoteChildReplaced(ChildReplacedNotification notification) =>
         SuppressNotificationForwarding(notification, () =>
         {
-            var localParent = Lookup(notification.Parent.GetId());
-            var replacement = (INode)notification.NewChild;
-            var newValue = ReplaceContainment(localParent, notification.Containment, notification.Index, replacement);
-
-            localParent.Set(notification.Containment, newValue, notification.NotificationId);
+            var newChild = LookupOpt(notification.NewChild.GetId()) ?? (INode)notification.NewChild;
```
erk-dik, line 221: Regarding `ChildReplacedNotification`:  Here I check existence of replacement node with LookupOpt, because I would think that replacement child (newchild) might already exists, but I think then it becomes move and replace type of notification, can you confirm this ? @enikao

enikao: I think we could remove _all_ `Lookup` calls from the replicator -- this should happen earlier in the pipeline (either the `DeltaXxxToNotificationMapper` or `NotificationToNotificationMapper`). Your call whether we want to change that in this PR.

erk-dik: I will handle this in another PR, this does not sound like a trivial change.

PR #147, opened 2025-10-08
introducing generic add/insert/remove node api for multiple Link features

erk-dik:


src/LionWeb.Core/Core/M1/M1Extensions.cs
```diff
@@ -48,14 +47,12 @@ public static void InsertBefore(this INode self, INode newPredecessor)
             // should not happen
             throw new TreeShapeException(self, "Cannot insert before a node in a single containment");
 
-        var list = new List<INode>(containment.AsNodes<INode>(enumerable));
-        var index = list.IndexOf(self);
+        var index = enumerable.AsNodes<INode>().GetIndexOf(node => node == self);
         if (index < 0)
```
erk-dik: This check looks redundant to me, unless `self `node is removed from the containment while InsertBefore/AfterBefore is still in execution. @enikao

enikao: It is (as mentioned in next line's comment), but I opt for defensive programming here -- this is called by users, and then we want to give proper Exception messages.

PR #150, opened 2025-11-10
bugfix - notification emitter considers whether a replacement node is part of a partition or not

erk-dik:
A floating node is a node which is not part of a partition. If  a node in a partition replaced by a floating node, then `ChildReplacedNotification `is emitter. If there is no replacement, then `ChildAddedNotification` is emitted.

### DIFF
--- a/test/LionWeb.Core.Test/Notification/Replicator/ReplicatorTests_Containment.cs
+++ b/test/LionWeb.Core.Test/Notification/Replicator/ReplicatorTests_Containment.cs
@@ -799,7 +799,7 @@ public class ReplicatorTests_Containment: ReplicatorTestsBase
     #endregion
 
     /// <summary>
-    /// A floating node is a node which is not part of any partition (in a forest).
+    /// A floating node is a node which is not part of any partition.
     /// In the test below, replacement node is a floating node,
     /// therefore; as a result of replacement, <see cref="ChildReplacedNotification"/> is emitted.  
     /// </summary>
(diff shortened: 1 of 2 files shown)


test/LionWeb.Core.Test/Notification/Replicator/ReplicatorTests_Containment.cs
```diff
@@ -798,5 +798,76 @@ public void ChildMovedInSameContainment_Backward()
...
+    }   
     
+    /// <summary>
+    /// Replaced and replacement nodes are part of a (same) partition.
+    /// Therefore <see cref="ChildMovedAndReplacedFromOtherContainmentNotification"/> ie emitted.
+    /// </summary>
+    [TestMethod]
+    public void ChildReplaced_nodes_are_in_same_partition()
+    {
+        var replaced = new Circle("replaced")
```
dslmeinte: Same comment as above on the names.

test/LionWeb.Core.Test/Notification/Replicator/ReplicatorTests_Containment.cs
```diff
@@ -798,5 +798,76 @@ public void ChildMovedInSameContainment_Backward()
...
+        Assert.IsInstanceOfType<ChildReplacedNotification>(notificationObserver.Notifications[0]);
+        AssertEquals([originalPartition], [clonedPartition]);
+    }   
     
+    /// <summary>
+    /// Replaced and replacement nodes are part of a (same) partition.
+    /// Therefore <see cref="ChildMovedAndReplacedFromOtherContainmentNotification"/> ie emitted.
+    /// </summary>
+    [TestMethod]
+    public void ChildReplaced_nodes_are_in_same_partition()
```
enikao: Let's also have tests moving between different partitions:

* One where both partitions are in the same forest
* One where the origin is in a different forest
* One where the origin is in no forest

The latter two probably emit a "wrong" notification. Document that, but have the tests active so we show we thought about that case and accept the limitation.

Issue #153, opened 2025-11-12
Replacing a node with equivalent, but not identical node creates superfluous "ChildReplaced" delta


A-Dalton:
example:

```
MyParent[id=p] {
  singleContainment = MyChild[id=c, c# object address=10] {
    value = "42"
  }
}
```

action:

```
MyParent.singleContainment = new MyChild[id=c, c# object address=20] {
  value = "42"
}
```

erk-dik:
Fixed.

PR #156, opened 2025-11-13
bugfix issue 153

erk-dik:
Fixes issue #153 

DeltaDeserializerHandler is made public and (if used) deserializer does skip node with id that appears both in deserialized nodes and dependent nodes.

### DIFF
--- a/src/LionWeb.Protocol.Delta/DeltaDeserializerHandler.cs
+++ b/src/LionWeb.Protocol.Delta/DeltaDeserializerHandler.cs
@@ -21,7 +21,7 @@ using Core.M1;
 
 /// <summary>
 /// This handler enables deserializer to accept node id that appears
-/// both in deserialized nodes and dependent nodes.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres _loosely_ to [Semantic Versioning](https://semver.org/s
 * Moved `WebSocketServer` and `WebSocketClient` to separate project in core repo for reuse.
 * Added generic add/insert/remove node api for multiple `link` (references and containments) features. 
 ### Fixed
+* Bugfix: DeltaDeserializerHandler is made public and and when used deserializer does skip node with id that appears both in deserialized nodes and dependent nodes.
```
enikao: * double "and"
* Please adjust the description both here and on `DeltaDeserializerHandler`: You're describing the same thing I can read from the method names, that's noise. Please describe _why_ we need to do this for deltas.

Issue #158, opened 2025-11-13
Adding several entries into optional multiple containment in same order as they are now leads to wrong notifications


A-Dalton:
Example:

```
myParent {
  children = [a, b, c, d]
}

myParent.AddChildren([a, b, c, d])
```

Should be a no-op, instead we get notifications with wrong indices (e.g. second notification is `ChildMovedInSameContainment(oldIndex: 1, newIndex: 3)` where it should be `ChildMovedInSameContainment(oldIndex: 0, newIndex: 3)`

PR #159, opened 2025-11-17
Check whether to be added nodes to multiple containment are equal to the available nodes in the containment or not

erk-dik:
Fixes issue #158

### DIFF
--- a/test/LionWeb.Core.Test/Notification/NoOperations/NoOpsTests_Containment.cs
+++ b/test/LionWeb.Core.Test/Notification/NoOperations/NoOpsTests_Containment.cs
@@ -45,8 +45,8 @@ public class NoOpsTests_Containment
         var a = new Circle("a");
         var b = new Circle("b");
         var c = new Circle("c");
-        var d = new Circle("c");
-        var originalPartition = new Geometry("a")
+        var d = new Circle("d");
+        var originalPartition = new Geometry("geo")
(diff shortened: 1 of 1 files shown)


test/LionWeb.Core.Test/Notification/NoOperations/NoOpsTests_Containment.cs
```diff
@@ -30,12 +30,52 @@ public void ChildReplaced_Single_sets_the_same_node()
...
         originalPartition.GetNotificationSender()!.ConnectTo(notificationObserver);
-        
+
         originalPartition.Documentation = documentation;
 
         Assert.AreEqual(0, notificationObserver.Count);
     }
+
+    [TestMethod]
+    public void ChildAdded_Multiple_adds_the_same_list_of_children()
```
dslmeinte: What about cases where e.g. `originalPartition.AddShapes([a, b, c]);`?

enikao:
Please change the PR title -- it will end up in git history, and I'd like to read the history without lookup up external bug descriptions.

test/LionWeb.Core.Test/Notification/NoOperations/NoOpsTests_Containment.cs
erk-dik: issue #160

PR #162, opened 2025-11-18
Resolve forward references

enikao:


src/LionWeb.Core/Core/BaseTypes.cs
```diff
@@ -1034,44 +1034,29 @@ protected PartitionInstanceBase(NodeId id) : base(id) { }
     protected internal abstract IPartitionNotificationProducer? GetNotificationProducer();
 }
 
-public interface IReferenceDescriptor
+public record ReferenceDescriptor(ResolveInfo? ResolveInfo, NodeId? TargetId, IReadableNode? Target)
```
erk-dik: does it make sense to have `ReferenceDescriptor `  and related code (extension methods) in a separate file ?it is a long file already.

src/LionWeb.Core/Core/BaseTypes.cs
```diff
@@ -1034,44 +1034,29 @@ protected PartitionInstanceBase(NodeId id) : base(id) { }
...
-        {
-            ReferenceDescriptor r => r,
-            { Target: null } => new ReferenceDescriptor(descriptor.ResolveInfo, descriptor.TargetId,
-                (R?)descriptor.Target),
-            { Target: R t } => new ReferenceDescriptor(descriptor.ResolveInfo, descriptor.TargetId, t),
-            _ => throw new InvalidValueException(null, descriptor)
-        };
-
-    public static IReferenceDescriptor FromNode(IReadableNode node) =>
+    public static ReferenceDescriptor FromNode(IReadableNode node) =>
```
erk-dik: this is not an extension method, maybe we can move to the type itself.

enikao:
@erk-dik @dslmeinte Which name do you like better for the `(resolveInfo, targetId, targetNode)` triple?

a) `ReferenceDescriptor`
b) `ReferenceTarget`
c) something else?

I'm undecided.

dslmeinte:
> > @erk-dik @dslmeinte Which name do you like better for the `(resolveInfo, targetId, targetNode)` triple?
> 
> > 
> 
> > a) `ReferenceDescriptor` b) `ReferenceTarget` c) something else?
> 
> > 
> 
> > I'm undecided.
> 
> 
> 
> I would go for b or  `ReferenceTargetDescriptor `?

`ReferenceTarget` should be in there, at the very least. `ReferenceTargetData`?

PR #172, opened 2025-11-27
handle the cases where received new nodes (via delta messages) are already exist in the model 

erk-dik:
If a notificaiton/delta event adds a new node, the node itself and its subnodes have to be new (not present in the model), if it is not the case we warn the user. For the replacement logic, we allow replacing if new node is both existing and to be replaced node,  after replacement model is valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres _loosely_ to [Semantic Versioning](https://semver.org/s
 
 ## [0.4.0] - tbd
 ### Added
+* Added the following check for a new node received via a delta event: If a delta event adds a new node, the node itself and its sub nodes have to be new (not present in the model) according to delta specification. If this is not the case, we warn the user by throwing an invalid notification exception. For the replacement logic, we allow replacing if a new node is a both existing and to be replaced node, we allow this because after replacement model is valid.  
 * Moved `WebSocketServer` and `WebSocketClient` to separate project in core repo for reuse.
 * Added generic add/insert/remove node api for multiple `link` (references and containments) features.
 * Added `UnresolvedReferencesManager` to heal unresolved references once the target becomes available.


enikao (commented):
Please adjust the title, it doesn't say anything, and it's wrong.

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,311 @@
...
+{
+    /// <summary>
+    /// According to spec, all descendants in an added node MUST be new (i.e. not present in the model).
+    /// This test case is legal in api but results in delta error because the existingChild is attempted to be added again
+    /// to the model 
+    /// </summary>
+    [TestMethod]
+    public void AddChildWithNewChildContainingExistingNode()
+    {
+        var existingChild = new LinkTestConcept("existingChild");
```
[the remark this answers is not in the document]
erk-dik: those tests get complex, AAA pattern might be a good idea to improve the readibility.

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -131,14 +135,59 @@ public void Receive(INotificationSender correspondingSender, INotification notif
...
+                }
+        }
+    }
+    
     #region Partitions
 
-    private void OnRemoteNewPartition(PartitionAddedNotification notification) =>
+    private void OnRemoteNewPartition(PartitionAddedNotification notification)
+    {
+        CheckIfNewNodeContainsExistingNodes(notification, []);
```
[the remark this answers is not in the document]
enikao: It follows our general pattern: If something is unrecoverable wrong in processing notifications, we throw.

The actual clients [reports this via an event](https://github.com/LionWeb-io/lionweb-csharp/blob/58ee9f346aef4d5772b61c52f4ea365c7fa78107/src/LionWeb.Protocol.Delta/Client/LionWebClientBase.cs#L87).

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -131,14 +135,59 @@ public void Receive(INotificationSender correspondingSender, INotification notif
...
+        var hasIntersection = _sharedNodeMap.NodeIds.ToHashSet().Overlaps(newNodes);
+
+        if (!hasIntersection)
+        {
+            return;
+        }
+
+        var existingNodes = _sharedNodeMap.NodeIds.Intersect(newNodes).ToList();
+
+        switch (newNodeNotification)
```
enikao: We don't need this switch: Just take whatever is inside `acceptableNewNodes` and work with it.

We don't _want_ to care about the specific notification here.
(Maybe for this reason, we should only pass in the new node, not the notification?)

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -131,14 +135,59 @@ public void Receive(INotificationSender correspondingSender, INotification notif
         }
     }
 
+    private void CheckIfNewNodeContainsExistingNodes(INewNodeNotification newNodeNotification, HashSet<NodeId> acceptableNewNodes)
```
enikao: `acceptableNewNodes` is not a great name: It contains _existing_ nodes that are also mentioned in _new_.

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,311 @@
...
+        {
+            changedPartition.Containment_1 = replacement;
+        });
+        
+        // changes applied to original partition
+        Assert.AreSame(replacement, changedPartition.Containment_1);
+        Assert.IsNull(existingParent.Containment_0_1);
+        
+        // change has not replicated to the clone
+        Assert.ThrowsExactly<UnsetFeatureException>(() => ((LinkTestConcept)clonedForest.Partitions.ToList()[1]).Containment_1);
```
enikao: Can't we check this with `Assert.False(TryGetContainment_1())`?

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,311 @@
...
+        
+        CreateDeltaReplicator(originalForest);
+        
+        Assert.ThrowsExactly<InvalidNotificationException>(() =>
+        {
+            originalForest.AddPartitions([partition]);
+        });
+    }
+    
+    //TODO: add tests with forest for replaced child case
```
enikao: What about that todo?

enikao (commented):
about the title/description: _how_ do we handle that case? _why_?

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
erk-dik: added

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -199,7 +237,15 @@ private void OnRemoteChildDeleted(ChildDeletedNotification notification) =>
             notification.Parent.GetPartition()?.GetNotificationProducer()?.ProduceNotification(notification);
         });
 
-    private void OnRemoteChildReplaced(ChildReplacedNotification notification) =>
+    private void OnRemoteChildReplaced(ChildReplacedNotification notification)
+    {
+        HashSet<NodeId> replacedNodes = M1Extensions
```
enikao: We have this code block now 3 times in the class, maybe worth factoring out.

src/LionWeb.Core/Core/Notification/Pipe/RemoteReplicator.cs
```diff
@@ -190,15 +242,25 @@ private void OnRemoteChildAdded(ChildAddedNotification notification) =>
...
+            .Descendants(notification.ReplacedChild, true, true)
+            .Select(node => node.GetId())
+            .ToHashSet();
+        
+        CheckIfNewNodeContainsExistingNodes(notification, replacedNodes);
```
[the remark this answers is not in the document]
erk-dik: added.

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        // Act
+        // This emits a ChildAddedNotification, "replacement" is a new node but its child "existingChild" is not.
+        // That's what makes the ChildAddedNotification "invalid".
+        var applyReplacement = () =>
+        {
+            originalPartition.Containment_1 = replacement;
+        };
+        
+        // Assert
+        Assert.ThrowsExactly<InvalidNotificationException>(applyReplacement);
```
enikao: I find it harder to read if we factor out the lambda.
In this case, act/assert fall a bit together -- nature of exceptions.

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        var clonedForest = CreateDeltaReplicator(originalForest);
+        
+        var replacement = new LinkTestConcept("replacement")
+        {
+            Containment_0_1 = existingChild
+        };
+
+        // Act
+        // add partition to the forest
+        originalForest.AddPartitions([originalPartition]);
```
enikao: still arrange

src/LionWeb.Core/Core/Notification/Partition/IPartitionNotification.cs
```diff
@@ -335,7 +335,7 @@ public record AnnotationAddedNotification(
     public override NodeId ContextNodeId => Parent.GetId();
 
     /// <inheritdoc />
-    public IReadableNode NewNode => NewNode;
+    public IReadableNode NewNode => NewAnnotation;
```
erk-dik: fixes a hidden bug (I suppose).

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        
+        var clonedForest = CreateDeltaReplicator(originalForest);
+
+        var replacement = new LinkTestConcept("replacement")
+        {
+            Containment_0_1 = existingChild
+        };
+        
+        // Act
+        originalForest.AddPartitions([originalPartition]);
```
enikao: still arrange

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        
+        // Assert
+        AssertEquals([originalPartition], [clonedPartition]);
+    }
+    
+    #endregion
+    
+    [TestMethod]
+    [Ignore("All nodes must have unique ids, not checked (yet) for partitions")]
+    public void AddExistingPartition()
```
enikao: We don't check this anywhere at the moment (e.g. adding multiple different children with the same id also succeeds).

Reason: That's only a small guarantee, as we also could have a duplicate id anywhere else in the model, and global checks are too expensive.

We could solve this globally by attaching a listener that checks uniqueness, and keeps track of known ids.

test/LionWeb.Protocol.Delta.Test/DeserializeExistingNodesTests.cs
```diff
@@ -0,0 +1,508 @@
...
+        
+        var clonedPartition = CreateDeltaReplicator(originalPartition);
+        
+        var replacement = new LinkTestConcept("replacement")
+        {
+            Containment_0_1 = existingChild
+        };
+        
+        // Act
+        originalForest.AddPartitions([originalPartition]);
```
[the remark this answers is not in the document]
erk-dik: there are 2 changes happening including adding partition, thats why I kept it in the Act section

Issue #193, opened 2026-04-01
Recommend Package Update


gragra33:
I would recommend updating [Utf8JsonAsyncStreamReader](https://github.com/gragra33/Utf8JsonAsyncStreamReader) to the latest version - there is a performance benefit.

enikao:
Thanks for the hint @gragra33 (-:

enikao:
Fixed in #195

PR #195, opened 2026-04-09
Fix lenient node detach from parent

### LINKED ISSUE #194: `DetachFromParent()` broken for LenientNode
Assume `LenientNode A` that *contains* `LenientNode B`. We call `B.DetachFromParent()`.

Expected result: `B.parent = null` and `A` doesn't contain `B` any more.

Actual result: `A` still contains `B`.

enikao:
Fixes #194, implements #193

### DIFF
--- a/src/LionWeb.Core/Core/LenientNode.cs
+++ b/src/LionWeb.Core/Core/LenientNode.cs
@@ -470,13 +470,13 @@ public class LenientNode : NodeBase, INode
         {
             switch (value)
             {
-                case IReadableNode:
+                case IReadableNode n when n == child:
                     RemoveFeature(key);
                     return true;
                 case IList list:
                     Index index = list.IndexOf(child);
                     if (index == -1)
-                        return false;
+                        continue;
 
                     list.RemoveAt(index);
                     if (list.Count == 0)
--- a/test/LionWeb.Core.Test/NodeApi/Dynamic/SetFeaturesTests.cs
+++ b/test/LionWeb.Core.Test/NodeApi/Dynamic/SetFeaturesTests.cs
@@ -18,6 +18,7 @@
 namespace LionWeb.Core.Test.NodeApi.Dynamic;
 
 using M3;
+using System.Collections;
 
 [TestClass]
 public class SetFeaturesTests:DynamicNodeTestsBase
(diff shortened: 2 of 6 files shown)


src/LionWeb.Core/Core/LenientNode.cs
```diff
@@ -470,13 +470,13 @@ protected override bool DetachChild(INode child)
...
             {
-                case IReadableNode:
+                case IReadableNode n when n == child:
                     RemoveFeature(key);
                     return true;
                 case IList list:
                     Index index = list.IndexOf(child);
                     if (index == -1)
-                        return false;
+                        continue;
```
dslmeinte, line 479: I find the semantics of a `return` much clearer (in this context) than that of `continue`.

dslmeinte (approved):
(but see question)

test/LionWeb.Core.Test/NodeApi/Dynamic/SetFeaturesTests.cs
```diff
@@ -18,6 +18,7 @@
 namespace LionWeb.Core.Test.NodeApi.Dynamic;
 
 using M3;
+using System.Collections;
 
 [TestClass]
 public class SetFeaturesTests:DynamicNodeTestsBase
```
erk-dik, line 24: Why this is called `SetFeaturesTests` and the other one `ReflectionTests` , I  see both tests use `Set`.

erk-dik (approved):

src/LionWeb.Core/Core/LenientNode.cs
enikao: `continue` goes into the next `foreach` loop, that's what we want.

test/LionWeb.Core.Test/NodeApi/Dynamic/SetFeaturesTests.cs
enikao: Good question ... no idea. Which name is clearer to you?

src/LionWeb.Core/Core/LenientNode.cs
dslmeinte: OK. I find `for`-loops i.c.w. both `return`s and a `continue` quite tedious to read.

Also: why not rename `key` to `feature` since it’s of type `Feature`?

test/LionWeb.Core.Test/NodeApi/Dynamic/SetFeaturesTests.cs
erk-dik: `SetFeaturesTests` is more explicit and clear (?)

src/LionWeb.Core/Core/LenientNode.cs
enikao: That's still not the same behavior: `break` instead of `return true` would yield to `return false`.

test/LionWeb.Core.Test/NodeApi/Dynamic/SetFeaturesTests.cs
enikao: actually, they are different: `SetFeatureTests` really only tests to set features in various ways, an `ReflectionTests` checks behavior of several reflection operations.

src/LionWeb.Core/Core/LenientNode.cs
dslmeinte: > That's still not the same behavior: `break` instead of `return true` would yield to `return false`.

Where am I suggesting a `break`?!

enikao: If there's no `return` c# requires a `break`.

dslmeinte: Again: I didn’t suggest changing anything. I just remarked that I find this style that mixes `return` and `continue` semantics tedious to interpret. Instead of thinking about that opinion, you essentially reply with “but `break` doesn’t work", effectively devaluating this PR.

PR #202, opened 2026-05-20
Enabled AOT-optimized JSON (de)serialization for delta messages

### LINKED ISSUE #201: Don't format delta message JSON
Currently, delta JSON is pretty-printed. We should not do that.

enikao:
fixes #201


erk-dik (approved):
We do not really test whether changes are made comply with AOT or not. We might consider using: 

```
<PropertyGroup>
  <IsAotCompatible>true</IsAotCompatible>
</PropertyGroup>
```

Probably during publishing certain warnings from compiler are checked for AOT compatibility as well.

enikao:
> We do not really test whether changes are made comply with AOT or not. We might consider using:
> 
> ```
> <PropertyGroup>
>   <IsAotCompatible>true</IsAotCompatible>
> </PropertyGroup>
> ```
> 
> Probably during publishing certain warnings from compiler are checked for AOT compatibility as well.

Thanks, nice find!

PR #212, opened 2026-06-11
MigrationDeserializerHandler created way too many M2 instances for unknown keys

enikao:


test/LionWeb.Core.Benchmark/SerializerBenchmark.cs
```diff
@@ -118,11 +118,28 @@ public void Serialize_String_Aot()
...
+
+            return result;
+        });
+    
+    public static IEnumerable<T> CreateNodes<T>(long count, Func<string, T?, T?, List<T>?, T> factory)
     {
-        LinkTestConcept? lastLine = null;
-        LinkTestConcept? lastCircle = null;
-        LinkTestConcept? lastCoord = null;
+        T? lastLine = default;
```
dslmeinte: Why use T-inspired identifiers?

enikao: Because in the migration benchmark, I'm using `LinkTestConcept` generated for different LW versions.

dslmeinte: Ah, no, I meant `lastLine`, `lastCircle`, and `lastCoord` as Trumpf-inspired. (I get the generic `T`s.)

enikao: Aah, yes -- adjustment of an older test, only changed what's necessary.

enikao: fixed

PR #223, opened 2026-07-17
Refactor usages of obsolete Subscribe API

erk-dik:
… improve consistency in assertions

### DIFF
--- a/test/LionWeb.Core.Test/NodeApi/Generated/Annotation/Listener/Single/InsertTests.cs
+++ b/test/LionWeb.Core.Test/NodeApi/Generated/Annotation/Listener/Single/InsertTests.cs
@@ -15,12 +15,12 @@
 // SPDX-FileCopyrightText: 2024 TRUMPF Laser SE and other contributors
 // SPDX-License-Identifier: Apache-2.0
 
-namespace LionWeb.Core.Test.NodeApi.Generated.Annotation.Listener.Single;
+using LionWeb.Core.Notification.Partition;
+using LionWeb.Core.Test.Languages.Generated.V2024_1.Shapes.M2;
+using LionWeb.Core.Test.Languages.Generated.V2024_1.TestLanguage;
+using LionWeb.Core.Test.Notification;
 
-using Core.Notification.Partition;
-using Languages.Generated.V2024_1.TestLanguage;
-using Languages.Generated.V2024_1.Shapes.M2;
-using Notification;
+namespace LionWeb.Core.Test.NodeApi.Generated.Annotation.Listener.Single;
 
 [TestClass]
 public class InsertTests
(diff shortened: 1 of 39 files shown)


test/LionWeb.Core.Test/NodeApi/Generated/Annotation/Listener/Single/InsertTests.cs
```diff
@@ -32,18 +32,15 @@ public void Empty()
...
-            Assert.AreEqual(0, args.Index);
-            Assert.AreEqual(bom, args.NewAnnotation);
-        });
+        var observer = new NotificationObserver();
+        parent.GetNotificationSender()!.ConnectTo(observer);
 
         line.InsertAnnotations(0, [bom]);
 
-        Assert.AreEqual(1, notifications);
+        var notifications = observer.OfType<AnnotationAddedNotification>(1);
```
enikao: How does this react if there's not exactly 1 notification at all?
We might want to keep that assert.

enikao (commented):
looking good

test/LionWeb.Core.Test/NodeApi/Generated/Annotation/Listener/Single/InsertTests.cs
erk-dik: If there are 0 or 2+ AnnotationAddedNotifications, the Assert.AreEqual(1, typed.Count, ...) inside will fail the test immediately with a descriptive message showing what was actually received.

enikao: And if there are other notifications?

erk-dik: original test does not also check other type of notifications. I took this approach for 1-to-1 mapping of the test cases.
I can also add check for total number of notifications ?

enikao: I think the original would fail if there were any other notifications?
Not 100 % sure about it, though.
In any case, let's keep the same level of assertions as before.

erk-dik: OfType<T>(count) is now strict — it asserts:

- Exactly count notifications of type T exist
- The total notification count equals count (no other notification types)

OfTypeAmong<T>(count) is the relaxed variant — it only asserts the typed count, allowing other notification types to coexist. Used for tests that legitimately expect mixed types (like MultiListeners_SomeRead and MultiListeners_AllRead).

test/LionWeb.Core.Test/Notification/NotificationListAssertExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+/// Extension methods for asserting on collected notifications from <see cref="NotificationObserver"/>.
+/// </summary>
+public static class NotificationListAssertExtensions
+{
+    /// <summary>
+    /// Asserts that <b>exactly</b> <paramref name="count"/> notifications were received, all of type <typeparamref name="T"/>,
+    /// and returns them as a typed list.
+    /// Fails if any other notification types are present.
+    /// </summary>
+    public static List<T> OfType<T>(this NotificationObserver observer, int count) where T : INotification
```
enikao: Can we name this `AssertOfType` or similar? `OfType` is easy to mix up with the built-in extension method on enumerables.
(same for next method)

enikao (approved):

PR #225, opened 2026-07-20
Replace usages of geometry lang with test lang

erk-dik:


erk-dik:
What's needed in TestLanguage to fully eliminate Shapes:

- An annotation with restricted annotates — e.g., an annotation that only annotates LinkTestConcept (not Node). This would allow testing "can't annotate this parent" scenarios.

- A concept with two single-optional containments of the same type — e.g., LinkTestConcept getting a second Containment_0_1_b feature, or a new concept with two such features. This enables OtherInSameInstance tests.

- A type usable as both annotation and containment child — like Shapes' Documentation which is an annotation but also accepted by Geometry.Documentation containment. This enables the annotation↔containment movement tests.

Shall we update TestLanguage to accomadate the changes above ? @enikao

enikao:
> A type usable as both annotation and containment child — like Shapes' Documentation which is an annotation but also accepted by Geometry.Documentation containment. This enables the annotation↔containment movement tests.

This is not allowed (anymore) in LionWeb, so let's ignore that case for now.

enikao:
> * An annotation with restricted annotates — e.g., an annotation that only annotates LinkTestConcept (not Node). This would allow testing "can't annotate this parent" scenarios.
> 
> * A concept with two single-optional containments of the same type — e.g., LinkTestConcept getting a second Containment_0_1_b feature, or a new concept with two such features. This enables OtherInSameInstance tests.

Sounds good to me. Leaves the hard part -- naming.

enikao:
Proposal:

* `RestrictedTestAnnotation` annotates `LinkTestConcept` implements `INamed`
* `LinkTestConcept` gets `otherContainment_0_1`

Would that work?

dslmeinte:
> Proposal:
> 
> * `RestrictedTestAnnotation` annotates `LinkTestConcept` implements `INamed`
> * `LinkTestConcept` gets `otherContainment_0_1`
> 
> Would that work?

But please make those additions in `lionweb-integration-testing`, as that’s the authoritative source of the `TestLanguage` language. (Specifically: https://github.com/LionWeb-io/lionweb-integration-testing/blob/main/ts/scripts/generate-test-language.ts)

erk-dik:
> > Proposal:
> > 
> > * `RestrictedTestAnnotation` annotates `LinkTestConcept` implements `INamed`
> > * `LinkTestConcept` gets `otherContainment_0_1`
> > 
> > Would that work?
> 
> But please make those additions in `lionweb-integration-testing`, as that’s the authoritative source of the `TestLanguage` language. (Specifically: https://github.com/LionWeb-io/lionweb-integration-testing/blob/main/ts/scripts/generate-test-language.ts)

Sure ! Then I will do changes in test language in a separate PR.

enikao (approved):

